PR #17, opened 2016-01-05
feat(configs): setup app configuration

philloooo:
may not be the best setup, suggestions? @MurphyMarkW

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ docs/_build/
 
 local_settings.py
 settings.yaml
+*.sq3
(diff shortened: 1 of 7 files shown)


MurphyMarkW:
This looks generally good. I do want to poking at it this afternoon before merging, though, if that's ok. The main thing I want to look at is side-effects of using the python import system for configuration management, since there is technically caching going on when importing modules. I'm not entirely sure how this affects things like reload behavior under apache, if at all.

philloooo:
Yeah I never check how Apache reload works, but the default behavior will reload the settings

MurphyMarkW:
Ack, sorry, forgot to respond to this after my testing. This seems to work fine, so I'd say go ahead and merge it.

PR #18, opened 2016-04-15
fix(request-json): use dict.get

philloooo:
handle request payload's json elements same way as in [alias](https://github.com/LabAdvComp/indexd/blob/develop/indexd/alias/blueprint.py#L110) blueprint, avoiding 500 caused by `KeyError`
r? @MurphyMarkW


MurphyMarkW:
:+1:

PR #20, opened 2016-04-18
fix(error-handling): handle 404 in cross bp

philloooo:
- handle NoRecordFound in cross blueprint
- remove print statement
r? @MurphyMarkW


MurphyMarkW:
+1

PR #21, opened 2016-08-24
Feat/url api

SKotekal:
r? @philloooo 

I had to spin up a VM and edit in vim (really only familiar with :w and :q) since I was getting WindowsError with unit tests. Maybe this is cause of some format issues?


philloooo:
yeah windows had some nasty wierd characters... I believe you should be able to tweak something in sublime to fix such problems?

SKotekal:
save with encoding UTF-8? with BOM? Perhaps some other encoding?

SKotekal:
@philloooo I tried a fix with `.gitattributes` and [line endings](https://help.github.com/articles/dealing-with-line-endings/), which got the number of differences down from ~2K to 1.5K. Not sure how to proceed from here.

philloooo:
switch to pr to master

PR #23, opened 2017-01-11
Dockerfile

zflamig:


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get update && apt-get install -y python-pip git python-dev postgresql postgresql-client libpq-dev apache2 libapache2-mod-wsgi \ 
+ && apt-get clean && apt-get autoremove \
+ && rm -rf /var/lib/apt/lists/*
+ADD . /indexd
+RUN cd /indexd && pip install .
+
+RUN mkdir -p /var/www/indexd/ && chmod 777 /var/www/indexd && cp /indexd/wsgi.py /var/www/indexd/wsgi.py && cp /indexd/bin/indexd /var/www/indexd/indexd
+
+RUN echo '<VirtualHost *:80>\n\
+    WSGIDaemonProcess indexd processes=1 threads=3 python-path=/var/www/indexd/:/usr/bin/python home=/var/www/indexd\n\
```
philloooo: Can we support passing alternative processes and threads when `docker run`?

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:15.04
```
philloooo: Any reason not to use the latest LTS which is 16.04?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+MAINTAINER CDIS <cdissupport@opensciencedatacloud.org>
+
+RUN apt-get update && apt-get install -y python-pip git python-dev postgresql postgresql-client libpq-dev apache2 libapache2-mod-wsgi \
```
philloooo: if we want to use postgres I think it will be a dedicated postgres server outside of the docker so there is no need to install any postgres packages here except `libpq-dev`?

PR #37, opened 2017-05-15
chore(docker): Updated docker to use alpine, nginx & uwsgi

zflamig:
Build time is now 30 seconds and image size 220mb vs 2 minutes 40 seconds & 562mb.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+.git
+.gitignore
+.travis.yml
+.dockerignore
+tests
+test-requirements.txt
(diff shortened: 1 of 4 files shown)


kyleaschmitt:
It constitutes an OS distribution change.  We need to do due diligence and evaluate it first.  But I'm heartily impressed with the speed and size improvement you managed with it.

zflamig:
@fauzigo Can you take a look at this when you get a chance and see if we an switch to Alpine based images?

PR #46, opened 2017-10-23
feat(record-type): add `purpose` table 

### LINKED ISSUE #41: add a `type` property to IndexRecordURL and another model to store controlled enum in db
we need a separate field to describe the purpose of the data that's from a list of types that make sense to the system.
This field is needed to make services able to infer the purpose of that storage location. eg:
- on_premises_cold_storage
- on_premises_primary_storage
- amazon_backup
- amazon_primary_storage

rudyardrichter:
Resolves #41.

Adds the `IndexRecordUrlPurpose` table, containing a single column `purpose` which defines an enumeration of possible types/purposes for the data.

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -1,4 +1,3 @@
-from enum import Enum
 import uuid
 
 from contextlib import contextmanager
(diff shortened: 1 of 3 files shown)


philloooo:
- the `purpose` column need to be exposed from the api for both create, update and get
- need a db migration script for adding the column - will just be an optional field - you can use sqlalchemy migration tool or just create a raw sql like [this](https://github.com/uc-cdis/userdatamodel/tree/master/migrations)
- need to update swagger doc for 1)

PR #53, opened 2017-11-14
Update swagger doc for supporting versioning

giangbui:


openapis/swagger.json
```diff
@@ -1,594 +1,794 @@
...
-              "$ref": "#/definitions/AliasInputInfo"
-            }
-          }
-        ],
-        "responses": {
-          "200": {
-            "description": "successful operation",
-            "schema": {
-              "$ref": "#/definitions/AliasRef"
+        "/index/version/{UUID}": {
```
philloooo: I think it would be better to not overload the second subdirectory to be either {UUID} or `version`, how about just update the existing index creation endpoint to allow an optional `base_id` property in the json payload?

openapis/swagger.json
```diff
@@ -1,594 +1,794 @@
...
-      "properties": {
-        "form": {
-          "type": "string",
-          "description": "how the entry was submitted to storage",
-          "enum": [
-            "object",
-            "container",
-            "multipart"
-          ]
+        "/index/version/latest/{UUID}": {
```
philloooo: same reasoning as my previuos comment, how about `/index/{UUID}/versions/latest`?

openapis/swagger.json
```diff
@@ -1,594 +1,794 @@
...
+                        ]
+                    }
+                ]
+            }
         },
-        "size": {
-          "type": "integer",
-          "format": "int64",
-          "description": "size in bytes"
+        "/index/version/all/{UUID}": {
```
philloooo: same as above, how about `/index/{UUID}/versions` ?

openapis/swagger.yaml
```diff
@@ -149,6 +149,90 @@ paths:
         - cdis_auth:
```
philloooo: let's remove this yaml file, should just autogenerate it from the json on client side and no need to keep in github

openapis/swagger.json
```diff
@@ -1,594 +1,794 @@
...
+                "baseid": {
+                    "type": "string",
+                    "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+                }
+            }
         },
-        "rev": {
-          "type": "string",
-          "pattern": "^[0-9a-f]{8}$"
+        "VersionOutputInfo": {
```
frickjack: As @philloooo suggests - if you augment the existing submission endpoint with a baseid in InputInfo and OutputInfo, then you can remove these Version* specific types which are redefining the type of objects in the index

frickjack:
Hey Giang - go ahead and delete openapis/swagger.json, and add a note to README.md that swagger.json can be generated from the swagger.yaml using the swagger-codegen tool:

* https://stackoverflow.com/questions/34733253/converting-a-swagger-yaml-file-to-json-from-the-command-line
* https://github.com/swagger-api/swagger-codegen/blob/master/README.md

Thanks for cleaning up my mess!  :-)

philloooo:
if you remove the json file, you can update `view in swagger`  link in [README](https://github.com/uc-cdis/indexd/blob/master/README.md) to use the yaml too

openapis/swagger.json
giangbui: Since the base_id can be NULL, how indexd can create a new version of the index record? If we want to update the existing index creation endpoint with optinal base_id in the json payload, all the index record should have non-NULL base_id

PR #57, opened 2017-11-22
Fix/size

### LINKED ISSUE #56: indexd index_record 'size' field need to be BIGINT
If we want to continue storing file size in bytes, then we need to update our schema:

class IndexRecord(Base):
    '''
    Base index record representation.
    '''
    __tablename__ = 'index_record'

    did = Column(String, primary_key=True)
    rev = Column(String)
    form = Column(String)
    size = Column(Integer)

------------
https://www.postgresql.org/docs/9.1/static/datatype-numeric.html

zflamig:
Make size a BigInteger, and support altering existing tables from int to bigint.

Resolves #56

### DIFF
--- a/indexd/alias/drivers/alchemy.py
+++ b/indexd/alias/drivers/alchemy.py
@@ -161,7 +161,7 @@ class SQLAlchemyAliasDriver(AliasDriverABC):
             
             record.name = name
             
-            if record.rev and rev != record.rev:
+            if rev is not None and record.rev and rev != record.rev:
                 raise RevisionMismatch('revision mismatch')
             
             if size is not None:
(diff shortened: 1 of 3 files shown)


indexd/alias/drivers/alchemy.py
```diff
@@ -83,9 +86,16 @@ def __init__(self, conn, **config):
         
         Base.metadata.bind = self.engine
         Base.metadata.create_all()
-        
         self.Session = sessionmaker(bind=self.engine)
 
+        md = MetaData()
```
philloooo: can we make a separate migration function and script instead of check then conditionally migrate for driver initialization? This also doesn't work on sqlite mode.

zflamig:
Solved in #60 instead.

PR #59, opened 2017-12-05
fix(alias): ignore if rev is not set

zflamig:
Rev is totally broken, but this is a quick patch for now.


philloooo:
either we remove revision from alias if we wanna make a design decision or we fix it on client side to check revision

PR #62, opened 2017-12-21
feat(file_name): add optional file name

philloooo:
- add an optional file name column
- add migration function


zflamig:
*sigh* lgtm

philloooo:
codacy complains about the concrete class's `ids` and `add_version` doesn't match method signature in the parent class, but it totally matches... so I am gonna force merge.

PR #66, opened 2018-01-03
fix(migration): Fix db migration

giangbui:
Fix db migration for the case of brand new db


philloooo:
why not detect whether it's an empty database in the beginning?
should update `init_schema_version`  to optionally be given a version number then in the driver init just 
```
if is_empty_database(self.engine):
  self.create_all()
  init_schema_version(
    driver=self, model=IndexSchemaVersion , version=CURRENT_SCHEMA_VERSION
  )
else if auto_migrate:
  self.migrate_index_database()
  ```
?

giangbui:
Just fixed it!

PR #73, opened 2018-01-22
feat(metadata): allow updating metadata

philloooo:
need to allow updating metadata

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -396,7 +396,9 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
 
         return ret
 
-    def update(self, did, rev, urls=None, file_name=None, version=None):
+    def update(self,
+               did, rev, urls=None, file_name=None,
+               version=None, metadata=None):
         '''
         Updates an existing record with new values.
         '''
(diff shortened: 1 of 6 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -416,11 +418,24 @@ def update(self, did, rev, urls=None, file_name=None, version=None):
...
+                    session.delete(url)
+
                 record.urls = [IndexRecordUrl(
                     did=record.did,
                     url=url
                 ) for url in urls]
 
+            if metadata is not None:
+                for md_record in record.index_metadata:
+                    session.delete(md_record)
```
kulgan: I feel like deleting all entries and creating new one with only what was supplied might lead to unexpected outcomes. For example say, I have an entry with a UUID stored somewhere else like in another db, having multiple meta data records. I might later want to update or set new field in the metadata without doing an indexd look up first. I would expect this to append to the existing entries instead of removing previous entries and leaving just this one. 
A way to do this will be to create a dict of the existing metadata entries before deleting and update the dict based on what was supplied before creating the metadata entries

philloooo: But then you have no way to delete/fix previous metadata record if the key is wrong, so I think updating the whole thing should be the first thing to support. If we want to support partial update, that should be a PATCH instead of PUT

PR #82, opened 2018-03-06
feat(hashes): Support Etag and CRC hash types

### LINKED ISSUE #81: Allow ETag and CRC hashes
S3 uses an md5 of individual part md5s followed by the number of parts as an ETag for a multipart upload. This should be supported as it is easily verified against S3.

dankolbman:
Adds etag and crc hash types.

resolves #81

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -28,6 +28,8 @@ ACCEPTABLE_HASHES = {
     'sha1': re.compile(r'^[0-9a-f]{40}$').match,
     'sha256': re.compile(r'^[0-9a-f]{64}$').match,
     'sha512': re.compile(r'^[0-9a-f]{128}$').match,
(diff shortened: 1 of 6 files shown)


indexd/index/blueprint.py
```diff
@@ -28,6 +28,8 @@
     'sha1': re.compile(r'^[0-9a-f]{40}$').match,
     'sha256': re.compile(r'^[0-9a-f]{64}$').match,
     'sha512': re.compile(r'^[0-9a-f]{128}$').match,
+    'crc': re.compile(r'^[0-9a-f]{8}$').match,
+    'etag': re.compile(r'^[0-9a-f]{32}(-\d+)?$').match
```
dankolbman: Similar functionality between alias and index really needs to be consolidated.

PR #86, opened 2018-03-14
chore(swagger): Update swagger file

giangbui:


openapis/swagger.yaml
```diff
@@ -116,6 +137,27 @@ paths:
...
             - 'write:index'
             - 'read:index'
+    post:
+      summary: >-
+        Add a new version of the index record associated with the given id
+      description: ''
+      operationId: addNewVersion
+      produces:
+        - application/json
+      parameters:
```
philloooo: need to include the body reference?

openapis/swagger.yaml
```diff
@@ -116,6 +137,27 @@ paths:
         - cdis_auth:
             - 'write:index'
             - 'read:index'
+    post:
+      summary: >-
+        Add a new version of the index record associated with the given id
```
philloooo: should add more description about what 'add a new version' means
explain that the content of the data is immutable, if you want to change the content, which means the hashes are changed, then it's a new index document with new uuid

openapis/swagger.yaml
```diff
@@ -116,6 +137,27 @@ paths:
...
+      description: ''
+      operationId: addNewVersion
+      produces:
+        - application/json
+      parameters:
+        - name: UUID
+          in: path
+          description: entry id
```
philloooo: this should be more descriptive too

openapis/swagger.yaml
```diff
@@ -90,6 +90,28 @@ paths:
         - cdis_auth:
             - 'write:index'
             - 'read:index'
+    get:
+      tags:
+        - index
+      summary: List all meta data records
```
philloooo: just say `list all records`?

PR #90, opened 2018-03-21
feat(swagger): generate swagger client

philloooo:
- fix migration script
- start some client tests to use swagger client
- fix some swagger client spec accordingly

blocked by https://github.com/uc-cdis/cdisutils-test/pull/12


philloooo:
codacy doesn't like me import fixture that's not called, but I have to do that to make pytest work

test-requirements.txt
```diff
@@ -1,2 +1,4 @@
 pytest==3.0.6
 pytest-flask==0.8.1
+-e git+https://github.com/uc-cdis/cdisutils-test.git@0.2.2#egg=cdisutilstest
+-e git+https://github.com/uc-cdis/indexclient.git@1.4.0#egg=cdisutilstest
```
jbarno: do you mean to overwrite the egg? Seems like nothing good could come of it.

tests/test_client.py
```diff
@@ -1,11 +1,13 @@
...
 import pytest
+from swagger_client.rest import ApiException
 
 from indexd.index.blueprint import ACCEPTABLE_HASHES
 
 
-def test_index_list(client):
-    assert client.get('/index/').status_code == 200
+def test_index_list(swg_client):
+    swg_client.list_entries()
```
jbarno: Is this just looking for an error? Doesn't look like a test.

tests/test_client.py
```diff
@@ -64,94 +64,84 @@ def test_index_create_with_multiple_hashes(client, user):
...
         'form': 'object',
         'size': 123,
         'urls': ['s3://endpointurl/bucket/key'],
         'hashes': {'md5': '8b9942cf415384b27cadf1f4d2d682e5'}}
 
-    assert client.post(
-        '/index/',
-        data=json.dumps(data),
-        headers=user).status_code == 400
+    try:
```
jbarno: I like the with.exception pattern that pytest gives better, but it doesn't matter.

test-requirements.txt
philloooo: typo

tests/test_client.py
philloooo: this is the same expectation as the previous assert, just that when the status_code is 200, swagger client doesn't raise exception

tests/conftest.py
```diff
@@ -36,3 +41,30 @@ def user(app):
...
+def swg_config(indexd_client): # noqa
+    config = swagger_client.Configuration()
+    config.host = indexd_client.url
+    config.username = indexd_client.auth[0]
+    config.password = indexd_client.auth[1]
+    yield config
+
+
+@pytest.fixture
+def swg_index_client(swg_config):
```
jbarno: Just curious. Why are these yielding, but take no cleanup actions?

PR #93, opened 2018-03-26
Added indexes for 4 tables (url, hash, metadata, url-metadata)

phuongph:


jbarno:
Shouldn't SCHEMA_VERSION_NUMBER have been updated to 5?
@philloooo

philloooo:
you are right... @jbarno

PR #94, opened 2018-03-26
Fixed schema version number

phuongph:


jbarno:
Is there ever a time when we will need to run two migrations per version number? Because then I think this could just be the `CURRENT_SCHEMA_VERSION= len(SCHEMA_MIGRATION_FUNCTIONS)` to eliminate this potential issue.

jbarno:
Also this is the thing I was thinking of: https://www.pythoncentral.io/migrate-sqlalchemy-databases-alembic/

philloooo:
Yeah let’s just do len(

PR #121, opened 2018-04-27
feat(latest_version): allow filter by has_version

### LINKED ISSUE #120: support getting latest version document that has version populated

philloooo:
resolve #120

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -703,11 +703,10 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
 
         return ret
 
-    def get_latest_version(self, did):
+    def get_latest_version(self, did, has_version=None):
         """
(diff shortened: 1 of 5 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -721,9 +720,11 @@ def get_latest_version(self, did):
                 raise MultipleRecordsFound('multiple records found')
 
             query = session.query(IndexRecord)
-            record = query.filter(IndexRecord.baseid == baseid) \
-                .order_by(IndexRecord.created_date.desc()).first()
-
+            query = query.filter(IndexRecord.baseid == baseid) \
```
jbarno: Isn't the first query unnecessary if has_version is true? Its just going to get overwritten.
Also kinda hard to tell with 5 assignments to `query` in one function

PR #128, opened 2018-05-16
urls_metadata partial match in list_with_params

khan08:
Couple of things that might need to change:
- use sqlalchemy contain method for url key
- pass urls_metadata:jsonString to GET url param

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -1,4 +1,5 @@
 import re
+import json
 import flask
 import jsonschema
 import os.path
@@ -75,11 +76,9 @@ def get_index():
     if size is not None and size < 0:
         raise UserError('size must be > 0')
 
-
     urls = flask.request.args.getlist('url')
 
     acl = flask.request.args.getlist('acl')
-
 
     file_name = flask.request.args.get('file_name')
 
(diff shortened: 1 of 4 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -346,6 +347,20 @@ def ids(self,
                         ))
                     query = query.filter(IndexRecord.did.in_(sub.subquery()))
 
+            if urls_metadata is not None and urls_metadata:
+                query = query.join(IndexRecord.urls).join(IndexRecordUrl.url_metadata)
+                for url_key, url_dict in urls_metadata.items():
+                    query = query.filter(IndexRecordUrlMetadata.url.contains(url_key))
```
khan08: Contain would check if url contains a substring that matches the passed url. No wildcard is allowed in this case. We can also use SQL LIKE and use % as wildcard.

indexd/index/blueprint.py
```diff
@@ -94,6 +93,9 @@ def get_index():
     metadata = flask.request.args.getlist('metadata')
     metadata = {k: v for k, v in map(lambda x: x.split(':', 1), metadata)}
 
+    urls_metadata = flask.request.args.get('urls_metadata')
+    urls_metadata = urls_metadata and json.loads(urls_metadata)
```
khan08: The urls_metadata is passed as a json string in the url param.

philloooo: catch if it's an invalid json

indexd/index/drivers/alchemy.py
khan08: Not validating urls_metadata schema here. If client pass in valid json string with invalid schema, 500 will be thrown. IndexClient should handle this?

PR #129, opened 2018-05-16
chore(baseid): add index

philloooo:
also refactor the code to not do 2 lookups for baseid and did.

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -4,7 +4,7 @@ import uuid
 
 from cdislogging import get_logger
 from contextlib import contextmanager
-from sqlalchemy import func, select, and_
+from sqlalchemy import func, select, and_, or_
 from sqlalchemy import String, Column, Integer, BigInteger, DateTime
 from sqlalchemy import ForeignKey, ForeignKeyConstraint, Index
 from sqlalchemy.orm import relationship, sessionmaker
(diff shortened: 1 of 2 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -509,18 +509,13 @@ def get(self, did):
         """
         with self.session as session:
             query = session.query(IndexRecord)
-            query = query.filter(IndexRecord.did == did)
+            query = query.filter(
+                or_(IndexRecord.did == did, IndexRecord.baseid == did)
+                ).order_by(IndexRecord.created_date.desc())
```
phuongph: Can you let a user flag here?
`def get(self, did, flag=True):`

`if flag:`
       `query = query.filter(or_ (...))`
`else:`
       `query = query.filter(IndexRecord.did == did)`

indexd/index/drivers/alchemy.py
```diff
@@ -905,9 +902,18 @@ def migrate_7(session, **kwargs):
...
 
+def migrate_8(session, **kwargs):
+    """
+    create index on IndexRecord.baseid
+    """
+    session.execute(
+        "CREATE INDEX ix_{tb}_baseid ON {tb} ( baseid )"
+        .format(tb=IndexRecord.__tablename__))
+
+
```
phuongph: Should check if `ix_index_record_baseid` index exist before create a new one

indexd/index/drivers/alchemy.py
phuongph: This can be very useful for my services

philloooo: I don't think we should keep adding flags for different tweaks unless it's necessary. We can merge this and if the performance is not good, we can think about adding flags later

indexd/index/drivers/alchemy.py
philloooo: None of the existing migration check that because it's relying on the migration version table

PR #138, opened 2018-06-19
feat(negate): allow negate query 

khan08:
- negate_params passed in as a whole json string.
- negate_params json string has the same format as doc json.
- if {key: value} key negated, return all docs that does not have the key at all.
- if {key: value} value negated, return all docs that don't have {key: value} at all or value unmatched.
- :arrow_up:  If return only docs that have the key but value unmatched, this will required runner to make multiple call to indexd to get node to backup.

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -109,7 +109,7 @@ def get_index():
         try:
             negate_params = json.loads(negate_params)
         except ValueError:
-            raise UserError('urls_metadata must be a valid json string')
+            raise UserError('negate_params must be a valid json string')
 
     records = blueprint.index_driver.ids(
         start=start,
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -76,6 +76,7 @@ def get_index():
     if size is not None and size < 0:
         raise UserError('size must be > 0')
 
+    # TODO: Based on indexclient, url here should be urls instead. Or change urls to url in indexclient.
     urls = flask.request.args.getlist('url')
 
     acl = flask.request.args.getlist('acl')
@@ -103,6 +104,13 @@ def get_index():
     if limit < 0 or limit > 1024:
         raise UserError('limit must be between 0 and 1024')
 
+    negate_params = flask.request.args.get('negate_params')
+    if negate_params:
+        try:
+            negate_params = json.loads(negate_params)
+        except ValueError:
+            raise UserError('urls_metadata must be a valid json string')
+
     records = blueprint.index_driver.ids(
         start=start,
         limit=limit,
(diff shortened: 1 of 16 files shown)


openapis/swagger.yaml
```diff
@@ -166,6 +166,13 @@ paths:
           description: comma delimited urls
           required: false
           type: string
+        - name: negate_params
+          in: query
+          description: |
+            all the negate params go in here as the same format with regular params
+            pass a Document as a json string with all the negated values.
```
philloooo: needs more explaination similar to what you have in the issue description

indexd/index/drivers/alchemy.py
```diff
@@ -371,6 +375,69 @@ def ids(self,
...
 
+    @staticmethod
+    def _negate_filter(session,
+                       query,
+                       urls=None,
+                       acl=None,
+                       file_name=None,
+                       version=None,
+                       metadata=None,
+                       urls_metadata=None):
```
philloooo: docstring to explain the negate behavior for list, dict and nested dict

indexd/index/blueprint.py
```diff
@@ -103,6 +104,13 @@ def get_index():
     if limit < 0 or limit > 1024:
         raise UserError('limit must be between 0 and 1024')
 
+    negate_params = flask.request.args.get('negate_params')
+    if negate_params:
+        try:
+            negate_params = json.loads(negate_params)
+        except ValueError:
+            raise UserError('urls_metadata must be a valid json string')
```
philloooo: wrong error message

openapis/swagger.yaml
```diff
@@ -166,6 +166,18 @@ paths:
...
           type: string
+        - name: negate_params
+          in: query
+          description: |
+            All the negate params go in here as the same format with regular params:
+            pass a Document as a json string with all the negated values.
+            For string (version, file_name), filter with value != <value>
+            For list (urls, acl), filter with doc that don't HAS <value>
+            For dict (metadata, urls_metadata). In each (key,value) pair:
+            - if (not value is True): then filter with key doesn't exist
```
philloooo: if (value is null)?

indexd/index/drivers/alchemy.py
```diff
@@ -371,6 +375,88 @@ def ids(self,
...
+        """
+        param_values passed in here will be negated
+
+        for string (version, file_name), filter with value != <value>
+        for list (urls, acl), filter with doc that don't HAS <value>
+        for dict (metadata, urls_metadata). In each (key,value) pair:
+        - if (not value is True): then filter with key doesn't exist
```
philloooo: if (not value is None)?

indexd/index/drivers/alchemy.py
```diff
@@ -371,6 +375,88 @@ def ids(self,
...
+        - if value is provided, then filter with value != <value> OR key doesn't exist
+
+        :param session: db session
```
philloooo: we use [google docstring style](https://github.com/uc-cdis/rfcs/blob/master/text/0003-style-guide.md#documentation)

PR #149, opened 2018-07-31
feat(hs): Add HS client

zflamig:


indexd/blueprint.py
```diff
@@ -85,6 +86,9 @@ def dist_get_record(record):
             elif indexd['type'] == "dos":
                 fetcher_client = DOSClient(baseurl=indexd['host'])
                 res = fetcher_client.get(record)
+            elif indexd['type'] == "hs":
```
philloooo: can you add a comment about what hs is

PR #150, opened 2018-08-02
fix(update_doc): allow setting version to None 

kulgan:
```
if version is not None:
    record.version = version
```

This behavior blocks being able to nullify the version field. This PR gets rid of this by simply setting it to whatever is coming from the request. Looking at 
https://github.com/uc-cdis/indexd/blob/415f7338845f3e3aa6398d368d85157caab97b50/indexd/index/blueprint.py#L275
there is a possibility that this field is intentionally not set (which resolves to None either way), hence removing the check above leads to unintended results. Which led me to a second solution which uses a sentinel (eg -1, @, ^ etc); which will imply that the user needs to know this special symbol not to use in their versioning scheme. I skipped this solution cos of that.

The implemented solution is in line with using the IndexClient, since to do an update you will need first do a get document, which will mean the right version would be on the document, unless set to something else before calling the patch

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -262,31 +262,19 @@ def post_index_record():
 @blueprint.route('/index/<path:record>', methods=['PUT'])
 @authorize
 def put_index_record(record):
-    '''
+    """
     Update an existing record.
-    '''
+    """
     try:
         jsonschema.validate(flask.request.json, PUT_RECORD_SCHEMA)
     except jsonschema.ValidationError as err:
         raise UserError(err)
-
     rev = flask.request.args.get('rev')
-    file_name = flask.request.json.get('file_name')
-    version = flask.request.json.get('version')
-    urls = flask.request.json.get('urls')
-    acl = flask.request.json.get('acl')
-    metadata = flask.request.json.get('metadata')
-    urls_metadata = flask.request.json.get('urls_metadata')
 
     did, baseid, rev = blueprint.index_driver.update(
         record,
         rev,
-        file_name=file_name,
-        version=version,
-        urls=urls,
-        acl=acl,
-        metadata=metadata,
-        urls_metadata=urls_metadata,
+        flask.request.json,
     )
 
     ret = {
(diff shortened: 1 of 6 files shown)


ProfOak:
The issue is only when a user is hitting the indexd api directly. When you update a document through IndexClient you have to first retrieve the doc then run `patch()`. The values don't change from the original get, unless a user requests it explicitly. So at that point it is explicit, but only through IndexClient.

What if we added a URL parameter on this endpoint that conveys the explicit consent to update the `version` attribute? Something along these lines:

```
POST /index?allow_none_versions=true
```

We could have this be enabled 100% of the time in the IndexClient, because it's still explicit the way `patch()` works.

What do you all think?

philloooo:
the indexclient is just one client implementation, however it behaves is not related to indexd, there are a lot of indexd users who don't use indexclient at all. So any indexd design decision shouldn't be affected by indexclient.

philloooo:
I don't think it makes sense to introduce special flags or special symbols. It should be generally intuitive and not hardcoded for just the `version` field.
If we want to support nullify a field, the intuitive behavior should be:
```
PATCH
{
  $VARIABLE_A: null,
  $VARIABLE_B: 'abc'
}
-> nullify VARIABLE_A
PATCH
{
  $VARIABLE_B: 'abc'
}
-> nothing happens to VARIABLE_A
```

kulgan:
I'm not too convinced adding a flag would be any smoother, and what if you want to do this for more fields? Figuring out if the user sent in a field should be easy by doing
```
flask.request.json.has_key('version')
```
But passing the resulting value to  https://github.com/uc-cdis/indexd/blob/415f7338845f3e3aa6398d368d85157caab97b50/indexd/index/blueprint.py#L281 will be cancel the effect of checking if there's a key unless I pass an `extra flag` signaling that there is a key.

```
did, baseid, rev = blueprint.index_driver.update(
        record,
        rev,
        file_name=file_name,
        version=(version, flask.request.json.has_key('version')),
        urls=urls,
        acl=acl,
        metadata=metadata,
        urls_metadata=urls_metadata,
)
```
Then in `def update(...`
```
version # (version, listed)
if version[1]:
     record.version = version[0]
```
This looks like a hack to me. but would achieve the goal of only updating when the user specifies a property

PR #155, opened 2018-08-15
fix(update): potentially out of order dicts

ProfOak:
An update may cause there to be a urls_metadata entry without
a urls entry depending on what gets updated first. This causes a schema
error. To avoid that, update each special field individually.

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -1,22 +1,41 @@
 import datetime
-from future.utils import iteritems
 import uuid
+from contextlib import contextmanager
 
 from cdislogging import get_logger
-from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -658,47 +678,51 @@ def update(self, did, rev, changing_fields):
...
+                    changing_fields['urls_metadata'],
+                    record,
+                    session,
+                )
 
-                # No special logic needed for other updates.
-                # ie file_name, version, etc
-                else:
+            for key, value in changing_fields.items():
+                if key not in special_fields:
```
philloooo: 'special_fields' sounds confusing, can you rename to composite_fields? and you need to add a comment to say urls_metadata need to be processed after urls

PR #159, opened 2018-09-07
fix(list): add test for list with limit

philloooo:
try to reproduce in test


indexd/index/drivers/alchemy.py
```diff
@@ -385,6 +385,10 @@ def ids(self,
             if negate_params:
                 query = self._negate_filter(session, query, **negate_params)
 
+            # joining url and url metadata will have duplicate results
+            if urls_metadata or negate_params:
```
singergr: Do we care about `urls` or `acl`? We also join on those, but the way the logic currently works, I don't think we can ever match more than one row per record for either of those (we do a logical AND, so we match one row if one URL is given and no rows if more than one is given), so it may not matter. I guess it is okay as long as we remember to fix this condition as stuff changes.

singergr:
I think this will be okay. We are already sorting by DID anyway, so the DISTINCT should be easy.

I also tried merging in the changes from https://github.com/uc-cdis/indexd/pull/160 and verified that adding `joinedload`s doesn't break this.

indexd/index/drivers/alchemy.py
philloooo: oh you are right. For the acl join we need distinct too, others are using subquery so we don't need to do distinct select

philloooo: ahh didn't finish reading your comment, you are right, there will never be duplicate entries... so nothing need to change

PR #160, opened 2018-09-13
fix(TT-800): use joinedload in GET /index

singergr:
Update `SQLAlchemyIndexDriver.ids()` to use joinedload. This fixes the N+1 select issue in `GET /index`.

I tried testing various query parameters and got the same results before and after this change. I also saw a speedup (0.23 seconds vs. 1.20 seconds) when querying for a list of 100 IDs. Querying by URLs metadata didn't get much faster; for that case, I think we're limited by way we filter and/or the table structure.

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -20,7 +20,7 @@ from sqlalchemy import (
 )
 from sqlalchemy.exc import IntegrityError, ProgrammingError
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import relationship, sessionmaker
+from sqlalchemy.orm import joinedload, relationship, sessionmaker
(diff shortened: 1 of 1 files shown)


ProfOak:
Did you have a chance to test setting the join in the database definitions? Most of the endpoints involve getting a whole document, which joins most of the tables anyways.

eg
```python
class Parent(Base):
    __tablename__ = 'parent'

    id = Column(Integer, primary_key=True)
    children = relationship("Child", lazy='joined')
```
http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html

singergr:
I did configure the joins like that when doing the performance testing, but for this PR, I decided to limit the changes to just this query so that less of indexd would be affected. I don't think it would break anything, but this seemed a little safer.

PR #162, opened 2018-09-14
feat(urls_endpoint): dedicated urls endpoint

kulgan:


indexd/index/drivers/query/urls.py
```diff
@@ -0,0 +1,75 @@
...
+        self.driver = alchemy_driver
+
+    def query_urls(self, exclude=None, include=None, only_versioned=True, offset=0, limit=1000):
+
+        with self.driver.session as sxn:
+            # special database specific functions dependent of the selected dialect
+            q_func = driver_query_map.get(sxn.bind.dialect.name)
+
+            query = sxn.query(IndexRecordUrl.did, q_func['string_agg'](IndexRecordUrl.url, ",").label("urls"))\
+                .outerjoin(IndexRecord)
```
khan08: I don't think outer join is needed here?

kulgan: Without it, you won;t be able to filter by versioned entries. Maybe I can add it only when (un)versioned entries are requested

khan08: oh right we always need the version column, I was thinking IndexRecord left outer join to IndexRecordUrl might have better performance, but on second thought I think it's the same..

philloooo:
can we move it to a separate root endpoint `/_query` ? the `/index/url` or `/index/url/metadata` is misleading because you would think the returned resource is url or url metadata but it's actually the index object.

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+        c_response = {}
+        if fields_dict.get("did"):
+            c_response["did"] = record[0]
+        if fields_dict.get("urls"):
+            c_response["urls"] = [record[1]]
+        if fields_dict.get("rev"):
+            c_response["rev"] = record[2]
+        results.append(c_response)
+
+    response = Response(json.dumps(results, indent=2, separators=(', ', ': ')), 200, mimetype="application/json")
```
rudyardrichter: for returning JSON use `flask.jsonify` and not `Response`

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+    results = []
+    for record in record_list:
+        c_response = {}
+        if fields_dict.get("did"):
+            c_response["did"] = record[0]
+        if fields_dict.get("urls"):
+            c_response["urls"] = record[1].split(",") if record[1] else []
+        results.append(c_response)
+
+    response = Response(json.dumps(results, indent=2, separators=(', ', ': ')), 200, mimetype="application/json")
```
rudyardrichter: use `flask.jsonify`

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@urls.record
+def pre_config(state):
+    driver = state.app.config["INDEX"]["driver"]
+    urls.logger = state.app.logger
+    urls.driver = AlchemyURLsQueryDriver(driver)
+
+
+def requested_fields(fields):
```
rudyardrichter: this is just keeping track of what fields are present it seems like? so it should really just be `set(fields.split(","))` and not using a dictionary

rudyardrichter: although it's not even necessary because flask can handle this directly

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if kwargs:
+        return Response("Unexpected Parameter(s) {}".format(kwargs), 400)
+
+    version = version.lower() in ["true", "t", "yes", "y"] if version else None
+    record_list = urls.driver.query_metadata_by_key(key, value, url=url,
+                                                    only_versioned=version,
+                                                    offset=int(offset), limit=int(limit))
+    fields = fields or "did,urls,rev"
+    fields_dict = requested_fields(fields)
```
rudyardrichter: this can just be `set(flask.request.args.get("fields", "").split(","))`

rudyardrichter: (and of course call it something other than `fields_dict`)

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+
+    results = []
+    for record in record_list:
+        c_response = {}
+        if fields_dict.get("did"):
```
rudyardrichter: now here instead of using `.get` just do `if "did" in fields_set` (or whatever)

indexd/index/drivers/query/urls.py
```diff
@@ -0,0 +1,75 @@
...
+            elif only_versioned is False:
+                query = query.filter(~IndexRecord.version.isnot(None))
+
+            # add url filter
+            if url:
+                query = query.filter(IndexRecordUrlMetadata.url.like("%{}%".format(url)))
+
+            # [('did', 'url', 'rev')]
+            record_list = query.order_by(IndexRecordUrlMetadata.did.asc()).offset(offset).limit(limit).all()
+        return record_list
```
rudyardrichter: it would be helpful to change this to a dictionary with the identifiers as keys, or put it into its own object, or something else rather than using `record_list[0]` etc. which is hard to understand

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+            `
+                [
+                    {"did": "AAAA-BB", "rev": "1ADs" "urls": ["s3://some-randomly-awesome-url"]},
+                    {"did": "AAAA-CC", "rev": "2Nsf", "urls": ["s3://another-randomly-awesome-url"]}
+                ]
+            `
+    """
+    # parameter validation
+    if kwargs:
+        return Response("Unexpected Parameter(s) {}".format(kwargs), 400)
```
rudyardrichter: don't need to use `Response` here; also this isn't consistent with the rest of the API, the error should be in JSON

rudyardrichter: also—I guess it's fine this way—but consider if just ignoring extra args is fine or if erroring is necessary

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+            c_response["urls"] = [record[1]]
+        if fields_dict.get("rev"):
+            c_response["rev"] = record[2]
+        results.append(c_response)
+
+    response = Response(json.dumps(results, indent=2, separators=(', ', ': ')), 200, mimetype="application/json")
+    return response
+
+
+@urls.errorhandler(UserError)
```
rudyardrichter: the app already has a `UserError` handler to translate things to 400s

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,115 @@
...
+    results = []
+    for record in record_list:
+        c_response = {}
+        if fields_dict.get("did"):
+            c_response["did"] = record[0]
+        if fields_dict.get("urls"):
+            c_response["urls"] = [record[1]]
+        if fields_dict.get("rev"):
+            c_response["rev"] = record[2]
+        results.append(c_response)
```
rudyardrichter: if you change the result from `query_metadata_by_key` then this chunk of logic isn't necessary and you can just go straight to returning the JSON

indexd/index/urls/v0.py
kulgan: I'll see see to formatting this response as json, I do think the extra args makes sense, it shows the args received and typos can quickly be resolved

indexd/index/urls/v0.py
kulgan: Yea, that failed  to parse the data. I can give it another go but no guarantees :)

indexd/index/drivers/query/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+        Args:
+            exclude (str): url pattern to exclude
+            include (str): url pattern to include
+            only_versioned (bool): query only versioned records or not
+            offset (int):
+            limit (int):
+        Returns:
+            list: result list
+        """
+        raise NotImplementedError("Not Implemented, use a concrete impl")
```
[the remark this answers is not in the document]
kulgan: There is nothing stopping one from extending a base class and not implementing the required methods. Your IDE will give you warnings if you use one, but that's it. Returning an error makes things more obvious at runtime, just doing a pass might just push the problem to some other section of the code

indexd/index/drivers/query/urls.py
```diff
@@ -0,0 +1,75 @@
...
+from indexd.index.drivers.query import URLsQueryDriver
+
+
+driver_query_map = {
+    "sqlite": dict(array_agg=func.group_concat, string_agg=func.group_concat),
+    "postgresql": dict(array_agg=func.array_agg, string_agg=func.string_agg)
+}
+
+
+class AlchemyURLsQueryDriver(URLsQueryDriver):
```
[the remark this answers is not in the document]
kulgan: I actually started out just extending `SQLAlchemy` driver. That driver class currently does everything and I do not want to inherit everything it does cos I don't need everything it does. So I decided to separate concerns which IMO makes things clearer to read and modify when functionality grows.

indexd/index/drivers/query/urls.py
kulgan: That would be nice, but sqlalchemy returns a list of sets, and so far I have not found away to get sqlalchemy to return a dict or custom object instead. In java I would dynamically transform the results to non entity objects. I also don't want to loop through the list here since another loop will be required to split the urls into an array and removed fields that should not be returned. I also find things like `result_list[0]` very confusing

indexd/index/urls/v0.py
kulgan: but using a dict is more efficient, and they both achieve the same result. Ideally, I would prefer not to do this second looping through the result, but I can't seem to get sqlalchemy return me the appropriate result structure

indexd/index/drivers/query/__init__.py
```diff
@@ -0,0 +1,37 @@
+from abc import ABCMeta, abstractmethod
+
+
+class URLsQueryDriver(object):
```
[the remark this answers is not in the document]
rudyardrichter: ah, my mistake. I didn't realize that only existed in 3

indexd/index/drivers/query/__init__.py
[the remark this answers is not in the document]
rudyardrichter: I mean, this actually will not run if a subclass doesn't implement every `abstractmethod`, thanks to `abc`. So the `raise` is unreachable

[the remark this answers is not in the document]
rudyardrichter:
```python
In [1]: from abc import ABCMeta, abstractmethod

In [2]: class Foo(object):
   ...:     __metaclass__ = ABCMeta
   ...:     @abstractmethod
   ...:     def test(self):
   ...:         pass
   ...:     

In [3]: class Bar(Foo):
   ...:     def __init__(self, x):
   ...:         self.x = x
   ...:         

In [4]: b = Bar()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-d6290cf35476> in <module>()
----> 1 b = Bar()

TypeError: Can't instantiate abstract class Bar with abstract methods test
```

indexd/index/__init__.py
```diff
@@ -0,0 +1,22 @@
+from functools import wraps
+from flask import request
+
+from indexd.errors import UserError
+
+
+def request_args_to_params(func):
```
rudyardrichter: Looking at this again I'm wondering why not just  use `flask.request.args`?

indexd/index/drivers/query/urls.py
```diff
@@ -0,0 +1,75 @@
...
+
+        with self.driver.session as sxn:
+            # special database specific functions dependent of the selected dialect
+            q_func = driver_query_map.get(sxn.bind.dialect.name)
+
+            query = sxn.query(IndexRecordUrl.did, q_func['string_agg'](IndexRecordUrl.url, ",").label("urls"))\
+                .outerjoin(IndexRecord)
+
+            # filter by version
+            if only_versioned is True:
```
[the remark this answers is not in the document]
rudyardrichter: Ok, thanks. I would probably name this `versioned` or something instead of `only_versioned` and add some documentation why it should be this way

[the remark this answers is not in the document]
rudyardrichter: (there's probably some clever way to do it in one statement with something from sqlalchemy but I'm not aware of a way to do it)

indexd/index/drivers/query/urls.py
rudyardrichter: right, so here couldn't you just return a dictionary?
```python
return dict(zip(['did', 'url', 'rev'], record_list))
```

indexd/index/urls/v0.py
rudyardrichter: `flask.jsonify` is more or less just wrapping `flask.Response` and `json.dumps` so if something is broken in that case, it's actually broken

indexd/index/urls/v0.py
rudyardrichter: the performance is essentially identical (especially for such a small collection); either way it doesn't matter if you change `query_metadata_by_key` to return a dictionary, then you can just skip this part

indexd/index/__init__.py
kulgan: Mainly cos of clarity in method parameters. I still use `flask.request.args`, but just in a single place

indexd/index/urls/v0.py
kulgan: beats me, I keep getting something like this when I run the tests using jsonify
` UserError: ("{'url': u'awesome-x', 'value': u'uploaded', 'key': u'state'}", ValueError('dictionary update sequence element #0 has length 3; 2 is required',)) `. Not sure why. but raw json.dumps works fine

indexd/app.py
```diff
@@ -4,6 +4,9 @@
 from .alias.blueprint import blueprint as indexd_alias_blueprint
 from .dos.blueprint import blueprint as indexd_dos_blueprint
 from .blueprint import blueprint as cross_blueprint
+
+from indexd.index.urls.v0 import urls as index_urls_blueprint
```
philloooo: syntactical consistency is important in general, can we make the import statement and file structure in the same fashion as other blueprints? same with other minor things like `with self.driver.session as session` instead of `with self.driver.session as sxn`

indexd/index/drivers/query/__init__.py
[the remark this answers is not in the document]
rudyardrichter: I agree, this is just the behavior of `abc`, like this snippet I posted above if you `raise` instead of `pass` it doesn't matter, it will never get there if the class inherits from an `ABC`.

I will go ahead and walk away from this bike shed now...

indexd/index/__init__.py
kulgan: The rest of the code already raises `NotImplementedError` for abstract methods, yet :). For this endpoints, Performance hit? I am not sure about that, but doing it this way, I can easily report parameter errors to back to the client without have to write much code, which I think is one of the benefits of python

indexd/app.py
rudyardrichter: I would say both—I think that's what @philloooo meant as well

indexd/index/urls/v0.py
```diff
@@ -0,0 +1,108 @@
...
+            `
+                [
+                    {"did": "AAAA-BB", "rev": "1ADs" "urls": ["s3://some-randomly-awesome-url"]},
+                    {"did": "AAAA-CC", "rev": "2Nsf", "urls": ["s3://another-randomly-awesome-url"]}
+                ]
+            `
+    """
+    # parameter validation
+    if kwargs:
+        return jsonify(dict(error="Unexpected query parameter(s)", params=kwargs.keys())), 400
```
rudyardrichter: this includes valid paramters in the error message

indexd/index/__init__.py
rudyardrichter: I guess the other code is a bad example. :slightly_smiling_face: 

Copying the values out from `flask.request.args` into another dict (the `kwargs`) is strictly speaking more costly than just using it directly, as is passing in extra arguments to the route functions. It's not a _dramatic_ performance hit, but I bring it up since you seemed concerned about the cost for a dictionary conversion from the URLs driver functions elsewhere. Getting the arguments out is still easy enough to do inline and if you want to actually return which parameters were unexpected you can't really avoid handling that separately.

rudyardrichter: I would suggest instead of messing with the route functions to just define separate functions that actually handle the computations, and in the routes just do this
```
return flask.jsonify(driver.query_urls(flask.request.args))
# let the function raise UserError or whatever, which gets handled automatically
```

indexd/app.py
kulgan: Cools, the urls and metadata are index queries, so I pictured them as sub packages under the index package, Will update accordingly

indexd/index/drivers/query/urls.py
kulgan: `urls` is a list, that the query return as a comma delimited string from the query, so doing dict -> zip still means I have to split it or convert it to list. The best solution here is to get sqlalchemy to work on it. If I move the logic in here, it will essentially be the same logic as is currently

indexd/index/__init__.py
kulgan: To be sure I follow, use of python's standard concept of decorators is expensive and possibly confusing, but I can replace the decorator with a method that does the exact same thing and add some code to detect invalid parameters?

kulgan: The sample you provided will raise an unexpected argument exception once there is a typo, which will skip any exception handling within the methods. I am also not worried about the performance here cos I know its just input parameters (much less in number) and I am not potentially looping over thousands of record as would be the case when manipulating the result.

kulgan: If you can, take a closer look at what using the decorator approach provides. Only unknown named parameters (probably typos or intentional invalid params) gets passed as **kwargs, and I am very sure it does not interfere with flask internal route system.

indexd/index/drivers/query/urls.py
rudyardrichter: yes, the thing is there's no reason to have it in this format because nothing is using it this way, only as a dictionary; it only gets used in `/metadata/q` which is doing the weird dictionary conversion right now, so it may as well be converted to a dictionary here, because it's clearer (one more note on this: you can also just `return [record.__dict__ for record in record_list]` which is probably the shortest/simplest that gets)

which `urls` are you talking about? I thought `IndexRecordUrlMetadata` had just the single `url` field?

indexd/index/__init__.py
rudyardrichter: you can use the same `**kwargs` method as now if you just pass `flask.request.args` to another function. the only difference is you don't need to put a decorator on the route functions to pass in the parameters as arguments.

rudyardrichter: suppose you change `query_metadata_by_key` to take a `**kwargs`, move the handling to there and raise `UserError` on unexpected arguments, and you can call `record_list = blueprint.driver.query_metadata_by_key(flask.request.args)`. does that make sense?

indexd/index/drivers/query/urls.py
rudyardrichter: I can't view it

kulgan: Expected response for for endpoints is like this:
```
[
 {"did": "AAAA-BB", "urls": ["s3://some-randomly-awesome-url"], "rev": "XXXX"},
 {"did": "AAAA-CC", "urls": ["s3://another-randomly-awesome-url"], "rev": "XXYX"}
]
```
Fields in this response can be removed based on the `fields` request parameter. 
SqlAlchemy returns a list of tuples. In `query_urls` the urls with that tuple is a comma separated list, in `query_urls_metadata` it started out as comma separated list of all the urls, but was reduced to just the url matching the pattern. To removed undesired columns and make the urls a list, I do one loop (which I am not happy with, but can't find an alternative using only sqlalchemy).
You are suggesting I do an initial loop, to convert result_list to a dict automatically. I can do that, but it would not cater for removing undesired fields and converting urls to a list. Meaning I either loop through the dict again, to do this or skip the dict conversion you suggested and bring in the current logic for removing the undesired fields and converting urls to list. If I do option 2, it will be the exact same solution as I currently have now and will still use something like result_lis[1]

PR #209, opened 2019-03-18
feat(rbac): add rbac field and API

fantix:
- [x] Add database field and migration
- [x] Add API field
- [x] Add unit tests
- [x] Add docs
- [ ] Add data migration

### New Features
- Record now have a new `rbac` field

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -78,6 +78,7 @@ class IndexRecord(Base):
     file_name = Column(String, index=True)
     version = Column(String, index=True)
     uploader = Column(String, index=True)
+    rbac = Column(String)
 
     urls = relationship(
         'IndexRecordUrl',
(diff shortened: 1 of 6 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -1175,9 +1180,16 @@ def migrate_10(session, **kwargs):
         .format(tb=IndexRecord.__tablename__))
 
 
+def migrate_11(session, **kwargs):
+    session.execute(
+        "ALTER TABLE {} ADD COLUMN rbac VARCHAR;"
+        .format(IndexRecord.__tablename__))
```
fantix: I was thinking to add this here:

```python
    for record in session.query(IndexRecord).options(joinedload(IndexRecord.acl)):
        if '*' in record.acl:
            record.rbac = '*'
        else:
            record.rbac = ' or '.join(record.acl)
```

But reading more about the user-sync logic, I feel like it might be just not as easy as above. So I'll create a new ticket to address that.

PR #211, opened 2019-04-22
bump sqlalchemy and sqlalchemy-utils; update .travis.yml

vpsx:
tldr: bump sqlalchemy and sqlalchemy-utils. Also specify jdk 8 in .travis.yml. 
------------------------

Tried to just bump sqlalchemy and sqlalchemy-utils. Travis error:
```
Package oracle-java8-set-default is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  oracle-java9-set-default
```
Probably because: https://launchpad.net/~webupd8team/+archive/ubuntu/java 

The Travis dist was `trusty` (just because that is the default)(...[until tomorrow!](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment)). Switching from java8 to java9 worked, as did doing 
```
before_install:
  - jdk_switcher use oraclejdk8
``` 
in trusty. But we prefer `xenial`(and it's the new default), and we don't really care about jdk 8 or 9 so long as Swagger works, so settled on specifying 
```
dist: xenial
jdk: 
    - oraclejdk8
```

### Dependency updates
bump sqlalchemy from 1.0.8 to 1.3.3
bump sqlalchemy-utils from 0.32.21 to 0.33.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
+dist: trusty
 language: python
 
 python:
   - "2.7"
   - "3.4"
-addons:
-  apt:
-    packages:
-      - oracle-java9-set-default
 
 sudo: false
 
+before_install:
+    - jdk_switcher use oraclejdk8
 install:
     - "wget https://oss.sonatype.org/content/repositories/releases/io/swagger/swagger-codegen-cli/2.3.1/swagger-codegen-cli-2.3.1.jar"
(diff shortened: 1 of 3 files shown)


fantix:
Sorry to comment on a closed PR: you probably need to update the versions in `setup.py` too.

PR #214, opened 2019-05-07
feat(authz): rename rbac to authz and use list

fantix:
### New Features
- added new authz field (previously rbac field)

### DIFF
--- a/indexd/index/driver.py
+++ b/indexd/index/driver.py
@@ -19,6 +19,8 @@ class IndexDriverABC(SQLAlchemyDriverBase):
             start=None,
             size=None,
             urls=None,
+            acl=None,
+            authz=None,
             hashes=None,
             file_name=None,
             version=None,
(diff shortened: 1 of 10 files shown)


indexd/index/driver.py
```diff
@@ -19,6 +19,8 @@ def ids(self,
             start=None,
             size=None,
             urls=None,
+            acl=None,
```
Avantol13: :grimacing: weird that acl wasn't here before

indexd/index/drivers/alchemy.py
```diff
@@ -801,6 +842,15 @@ def update(self, did, rev, changing_fields):
                     for ace in set(changing_fields['acl'])
                 ]
 
+            if 'authz' in changing_fields:
+                for resource in record.authz:
+                    session.delete(resource)
```
Avantol13: I assume `commit()` gets called somewhere outside this function?

indexd/index/driver.py
fantix: Yeah, it was an abandoned overdesign I think.

indexd/index/drivers/alchemy.py
fantix: Yes, `with self.session as session` does that I believe.

PR #218, opened 2019-05-28
chore(authz-migration): add migration script

rudyardrichter:
### Deployment changes

- Add script to run for filling out records' `authz` fields based on the `acl` fields.

### DIFF
--- a/bin/migrate_acl_authz.py
+++ b/bin/migrate_acl_authz.py
@@ -56,6 +56,8 @@ def main():
 
     with driver.session as session:
         q = session.query(IndexRecord)
+        if getattr(args, "start_did"):
+            q = q.filter(IndexRecord.did >= args.start_did)
         for record in windowed_query(session, q, IndexRecord.did, int(args.chunk_size)):
             if not record.acl:
                 logger.info(
--- a/bin/migrate_acl_authz.py
+++ b/bin/migrate_acl_authz.py
@@ -162,26 +162,35 @@ class ACLConverter(object):
             acl_item = acl_item.lstrip("u'")
             if acl_item != "*":
                 acl_item = re.sub(r"\W+", "", acl_item)
-            if acl_item == "*":
+
+            # update path based on ACL entry
+            if not acl_item:
+                # ignore empty string
+                continue
+            elif acl_item == "*":
+                # if there's a * it should just be open. return early
                 path = "/open"
                 break
             elif (
                 not self.use_sheepdog_db
                 or (projects_found == 0 and self.is_program(acl_item))
             ):
+                # if we don't have sheepdog we have to assume everything is a "program".
+                # also, we only want to set the path to a program if we haven't found a
+                # path for a project already.
                 path = "/programs/{}".format(acl_item)
                 programs_found += 1
-            else:
-                if not acl_item:
-                    return None
-                if acl_item not in self.projects:
-                    raise EnvironmentError(
-                        "program or project {} does not exist".format(acl_item)
-                    )
+            elif acl_item in self.projects:
+                # always want to update to project if possible
                 path = "/programs/{}/projects/{}".format(
                     acl_item, self.projects[acl_item]
                 )
                 projects_found += 1
+            else:
+                # nothing worked, raise exception
+                raise EnvironmentError(
+                    "program or project {} does not exist".format(acl_item)
+                )
 
         if not path:
             logger.error(
--- a/bin/migrate_acl_authz.py
+++ b/bin/migrate_acl_authz.py
@@ -56,9 +56,14 @@ def main():
 
     with driver.session as session:
         q = session.query(IndexRecord)
-        if getattr(args, "start_did"):
-            q = q.filter(IndexRecord.did >= args.start_did)
-        for record in windowed_query(session, q, IndexRecord.did, int(args.chunk_size)):
+        wq = windowed_query(
+            session,
+            q,
+            IndexRecord.did,
+            int(args.chunk_size),
+            start=getattr(args, "start_did")
+        )
+        for record in wq:
             if not record.acl:
                 logger.info(
                     "record {} has no acl, setting authz to empty"
@@ -249,7 +254,7 @@ class ACLConverter(object):
         return self.arborist_resources[path]
 
 
-def column_windows(session, column, windowsize):
+def column_windows(session, column, windowsize, start=None):
 
     def int_for_range(start_id, end_id):
         if end_id:
@@ -262,6 +267,8 @@ def column_windows(session, column, windowsize):
         .query(column, func.row_number().over(order_by=column).label('rownum'))
         .from_self(column)
     )
+    if start:
+        q = q.filter(column >= start)
     if windowsize > 1:
         q = q.filter(sqlalchemy.text("rownum %% %d=1" % windowsize))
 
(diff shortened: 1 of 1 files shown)


bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,191 @@
...
+        connection.close()
+        return
+
+    def is_program(self, acl_item):
+        return acl_item in self.programs
+
+    def acl_to_authz(self, record):
+        path = None
+        for acl_object in record.acl:
+            acl_item = acl_object.ace
```
Avantol13: should we do further sanitizing like stripping whitespace and maybe even stripping `u'` from front and `'` from back for problematic acls?

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,191 @@
...
+            if acl_item == "*":
+                path = "/open"
+            elif not path and self.is_program(acl_item):
+                path = "/programs/{}".format(acl_item)
```
Avantol13: I think we should namespace all these in `/gen3/programs` or something similar

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,191 @@
...
+                        "program or project {} does not exist".format(acl_item)
+                    )
+                path = "/programs/{}/projects/{}".format(
+                    acl_item, self.projects[acl_item]
+                )
+
+        if not path:
+            logger.error(
+                "couldn't get `authz` for record {} from {}; setting as empty"
+                .format(record.did, record.acl)
```
Avantol13: super nitpicky and maybe unnecessary/ugly, but if `record` that gets passed in doesn't have a `did` attribute this will die and not return. could use `getattr` and default to `unknown` or something for a safer log :man_shrugging:

bin/migrate_acl_authz.py
rudyardrichter: worth a shot :+1:

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,191 @@
...
+        for record in records:
+            if not record.acl:
+                logger.info(
+                    "record {} has no acl, setting authz to empty"
+                    .format(record.did)
+                )
+                record.authz = []
+                continue
+            try:
+                record.authz = acl_converter.acl_to_authz(record)
```
Avantol13: could perhaps skip records that already have authz field filled out? I'm thinking of a situation where this script fails halfway through and want to continue where it left off

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,191 @@
...
+                resource = {"path": path}
+                response = requests.post(url, timeout=5, json=resource)
+            except requests.exceptions.Timeout:
+                logger.error(
+                    "couldn't hit arborist to look up resource (timed out): {}".format(url)
+                )
+                failed = True
+            tag = None
+            try:
+                if response.status_code == 409:
```
Avantol13: I think you need to check `failed` before this. if it times out then `response` may not exist and this will die

bin/migrate_acl_authz.py
rudyardrichter: perhaps we should start with adding a namespace to the manifest globals, and use that here?

bin/migrate_acl_authz.py
rudyardrichter: good catch :ok_hand:

bin/migrate_acl_authz.py
Avantol13: yeah, we could do that. and you can get it into the deployment as an env var like: https://github.com/uc-cdis/cloud-automation/blob/2cbb29f2e40900ba76d702d493bf72f57f3a0a6c/kube/services/arborist/arborist-deploy.yaml#L104-L108

bin/migrate_acl_authz.py
rudyardrichter: I think it's safe to assume all records will have a `did`, since it's the primary key I don't think you could be returning records from a query without it

bin/migrate_acl_authz.py
rudyardrichter: I'm thinking it should go through the whole thing anyways, that would let us reset the authz fields if necessary, and for the cases where it's already cached the resource from arborist the time to do the conversion is minimal, just some string manipulation

rudyardrichter: Something I am concerned about is that with the current implementation, it doesn't commit anything until everything is done, so if the job gets killed early then it has to go through the whole thing again. Maybe commit every N records or something?

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,199 @@
...
+        logger.info("Can't import local_settings, import from default")
+        from indexd.default_settings import settings
+    driver = settings["config"]["INDEX"]["driver"]
+    try:
+        acl_converter = ACLConverter(args.sheepdog, args.arborist)
+    except EnvironmentError:
+        logger.error("can't continue without database connection")
+        sys.exit(1)
+    with driver.session as session:
+        records = session.query(IndexRecord)
```
Avantol13: what's going to happen if there are almost 2 million records? DCF prod has like 1.7 million. do we need to stream them through differently in chunks or something?

bin/migrate_acl_authz.py
Avantol13: that's a good point, and made me think of load handling for this, since DCF has over a million records. But yes, I agree, we should commit every so often

bin/migrate_acl_authz.py
Avantol13: oh wait, maybe this is some fancy generator ? `session.query`

rudyardrichter: to document the current discussion: looking into options for performant updates on a large number of records, maybe using something like [this](https://github.com/sqlalchemy/sqlalchemy/wiki/WindowedRangeQuery)

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    parser.add_argument(
+        "--arborist-url", dest="arborist", help="URL for the arborist service"
+    )
+    parser.add_argument(
+        "--chunk-size", dest="chunk_size", type=int, default=1000,
+        help="number of records to process at once",
+    )
+    parser.add_argument(
+        "--start-did", dest="start_did",
```
Avantol13: I don't think this gets used anywhere

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,264 @@
...
+
+        if self.namespace:
+            path = self.namespace + path
+
+        if path not in self.arborist_resources:
+            # add `?p` to create parent resources as necessary
+            url = "{}/resource/?p".format(self.arborist_url)
+            try:
+                resource = {"path": path}
+                response = requests.post(url, timeout=5, json=resource)
```
Avantol13: so this will not recreate existing resources and change the tag, right?

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if not path:
+            logger.error(
+                "couldn't get `authz` for record {} from {}; setting as empty"
+                .format(record.did, record.acl)
+            )
+            return None
+
+        if self.namespace:
+            path = self.namespace + path
```
Avantol13: we should ensure namespace starts with `/`

bin/migrate_acl_authz.py
```diff
@@ -0,0 +1,279 @@
...
+    return
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--path", default="/var/www/indexd/", help="path to find local_settings.py",
+    )
+    parser.add_argument(
+        "--sheepdog-db", dest="sheepdog", help="URI for the sheepdog database"
```
Avantol13: you can default this to `True` instead of doing a getattr() later, either way works

Avantol13: oh wait, I think I misunderstood this. it's the db connection string :man_facepalming: nix that then

PR #225, opened 2019-07-01
chore(sec): add security scanning pipeline

m0nhawk:
Added Jenkins security scan pipeline


.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+.gitignore export-ignore
+.travis.yml export-ignore
+Dockerfile export-ignore
+Jenkinsfile export-ignore
+Jenkinsfile.security export-ignore
+LICENSE export-ignore
+NOTICE export-ignore
+README.md export-ignore
+TestDockerfile export-ignore
+bin export-ignore
```
Avantol13: the bin contains python files we should probably scan

.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+deployment export-ignore
+# indexd
+openapis export-ignore
+pull_request_template.md export-ignore
+requirements.txt export-ignore
+setup.cfg export-ignore
+# setup.py
+test-requirements.txt export-ignore
+tests export-ignore
+wsgi.py export-ignore
```
Avantol13: we should probably scan this python file too

Jenkinsfile.security
```diff
@@ -0,0 +1,6 @@
+#!groovy
```
Avantol13: how does this file get called? not for every PR, right?

m0nhawk: You can check the actual pipeline in `cdis-jenkins-lib`.

It will run every Saturday: https://github.com/uc-cdis/cdis-jenkins-lib/blob/master/vars/securityPipeline.groovy#L14-L15

And it will skip scan every two Saturdays: https://github.com/uc-cdis/cdis-jenkins-lib/blob/master/vars/securityPipeline.groovy#L21-L42

PR #230, opened 2019-07-29
fix(tests): remove swagger-codegen dependecy

callenmacphee:
Pull request for swagger-codegen language removal from testing to get more accurate reading on coveralls. Conftest has been stripped of swagger_client fixtures. Tests previously dependent on swagger_client fixtures now depend on client and user fixtures. 

### Improvements
- Removed swagger-codegen dependency from tests; coverage report is now 80


tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
-    r1 = swg_index_client.update_blank_entry(r1.did, rev=r1.rev, body=updated)
-    r1 = swg_index_client.update_entry(
-        r1.did, rev=r1.rev, body={"acl": ["read"], "authz": ["read"]}
-    )
-    r1 = swg_index_client.get_entry(r1.did)
-    assert r1.acl == ["read"]
-    assert r1.authz == ["read"]
-    assert r1.did == did1
+    did_1 = rec_1["did"]
+    res_2 = client.put("/index/blank/" + did_1, headers=user, json=updated)
```
paulineribeyre: you are missing parameter `rev` here, but the test is passing because you didn't check the status code

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
+    rec_2 = res_2.json
```
paulineribeyre: i don't think this variable is used anywhere

tests/test_client.py
```diff
@@ -297,42 +365,57 @@ def test_get_empty_acl_authz_record_after_fill_size_n_hash(swg_index_client):
 
     # create the second blank record, only update size hashes and urls
```
paulineribeyre:
```suggestion
    # create the third blank record, only update size hashes and urls
```

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
+    assert rec_1["did"] == did_1
 
     # create the second blank record, only update size hashes and urls
     doc = {"uploader": "uploader_123"}
-    r2 = swg_index_client.create_blank_entry(doc)
-    did2 = r2.did
+    res_2 = client.post("/index/blank/", json=doc, headers=user)
+    rec_2 = res_2.json
+    did_2 = rec_2["did"]
     updated = {
```
paulineribeyre: this record is never actually updated even though the comment says it should be..? 🤔 can you add `client.put /index/blank/` for this record too?

tests/test_client.py
```diff
@@ -297,42 +365,57 @@ def test_get_empty_acl_authz_record_after_fill_size_n_hash(swg_index_client):
...
-    assert len(r.records) == 2
-    assert {r.records[0].did, r.records[1].did} == {did2, did3}
+    res = client.get("/index/?uploader=uploader_123&acl=null")
+    rec = res.json
+    assert len(rec["records"]) == 2
+    assert {rec["records"][0]["did"], rec["records"][1]["did"]} == {did_2, did_3}
 
 
-def test_urls_metadata(swg_index_client):
+def test_urls_metadata(client, user):
```
paulineribeyre:
```suggestion
def test_update_urls_metadata(client, user):
```

tests/test_client.py
```diff
@@ -462,55 +559,67 @@ def test_delete_and_recreate(swg_index_client):
...
+    )
+
+    # make sure it's deleted
+    res = client.get("/index/" + old_record["did"])
+    assert res.status_code == 404
 
     # create new doc with the same baseid and did
-    new_result = swg_index_client.add_entry(new_data)
+    new_result = client.post("/index/", json=new_data, headers=user)
+    print(new_result.status_code)
```
paulineribeyre: assert instead of print

tests/test_client.py
```diff
@@ -536,104 +647,115 @@ def test_index_create_with_duplicate_acl_authz(swg_index_client):
...
-    assert res.urls_metadata == {}
-    assert res.size is None
-    assert res.version is None
-    assert res.urls_metadata == {}
+    res = client.post("/index/blank/", json=doc, headers=user)
+    rec = res.json
+    assert rec["did"]
+    res = client.get("/index/" + rec["did"])
+    rec = res.json
+    assert rec["acl"] == []
```
paulineribeyre: could use `assert_blank` here

tests/test_client.py
```diff
@@ -536,104 +647,115 @@ def test_index_create_with_duplicate_acl_authz(swg_index_client):
...
-    assert r.version == data["version"]
+    res = client.post("/index/", json=data, headers=user)
+    rec = res.json
+    res = client.get("/index/" + rec["did"])
+    rec = res.json
+    assert rec["version"] == data["version"]
 
 
-def test_index_create_blank_record(swg_index_client):
+def test_index_create_blank_record(client, user):
```
paulineribeyre:
```suggestion
def test_create_blank_record_with_baseid(client, user):
```
to match tests `test_create_blank_record` and `test_create_blank_record_with_file_name`

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
+    assert rec["did"]
+    assert rec["rev"]
+    assert rec["baseid"]
+
+    res = client.get("/index/?uploader=uploader_123")
+    rec = res.json
+    assert rec["records"][0]["uploader"] == "uploader_123"
+    assert not rec["records"][0]["file_name"]
+
+    # statements below replace assert_blank
```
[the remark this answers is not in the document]
callenmacphee: rec in this case is a dictionary I believe and assert blank isn't formatted for that. I can change the definition of assert blank but I was afraid there might be some other dependencies I may mess up?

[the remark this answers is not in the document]
paulineribeyre: assert_blank is only used in the tests

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
     """
     Test that can fill size and hashes for empty record
     """
     doc = {"uploader": "uploader_123"}
 
-    r = swg_index_client.create_blank_entry(doc)
-    assert r.did
-    assert r.rev
+    res = client.post("/index/blank/", json=doc, headers=user)
+    print(res.status_code)
```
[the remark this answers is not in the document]
callenmacphee: I'm get a 201 status code, is that correct? Response looks like this when printed:
<<class 'pytest_flask.plugin.JSONResponse'> streamed [201 CREATED]>

The rest of the test is running correctly, so I assume 201 is okay?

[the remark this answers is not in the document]
paulineribeyre: yes, 2xx indicates success, 201 specifically indicates a successful creation

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
+    data_list_all = data_by_hash.json
+    ids = [record["did"] for record in data_list_all["records"]]
+    assert rec_1["did"] in ids
+    assert rec_2["did"] in ids
 
-    ids = [record.did for record in r.records]
-    assert r_1.did in ids
-    assert r_2.did in ids
+    idslist = ",".join(ids)
+    data_by_ids = client.get("/index/?ids=" + idslist)
```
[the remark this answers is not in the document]
callenmacphee: codacy is complaining about this change, saying: 
"Prohibits string format operations that have more arguments than conversion specifiers declared."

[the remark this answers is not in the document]
paulineribeyre: my bad, it should be `"/index/?ids={}".format(",".join(ids))`

it worries me that this test is passing even with this mistake. it's because using `?ids=` without a value just returns all the records. could you fix this test so that the list of IDs we're querying is only a subset of all the existing IDs?

tests/test_client.py
```diff
@@ -27,268 +25,338 @@ def get_doc(
...
-    r1 = swg_index_client.get_entry(r1.did)
-    assert r1.acl == ["read"]
-    assert r1.authz == ["read"]
-    assert r1.did == did1
+    did_1 = rec_1["did"]
+    res_2 = client.put("/index/blank/" + did_1, headers=user, json=updated)
+    rec_2 = res_2.json
+    rev = rec_1["rev"]
+    body = {"acl": ["read"], "authz": ["read"]}
+    res_1 = client.put("/index/" + did_1 + "?rev=" + rev, headers=user, json=body)
```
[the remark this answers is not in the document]
callenmacphee: Should I change the syntax for strings in which there is only a single append?
for example, should 
res = client.get("/index/" + did)      -->      res = client.get("/index/{}".format(did))
?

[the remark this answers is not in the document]
paulineribeyre: nah, that's fine

tests/test_client.py
callenmacphee: Im going to change this test a little bit because I realized the baseid aspect wasn't really tested beyond the status code assertion, and it will also allow me to use assert_blank correctly.

tests/test_client.py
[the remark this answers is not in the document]
callenmacphee: To do this I queried for only one of the two posted ids

tests/test_client.py
```diff
@@ -819,15 +810,10 @@ def test_create_blank_record_with_baseid(client, user):
     assert res.status_code == 201
     rec = res.json
     assert rec["did"]
-    res = client.get("/index/" + rec["did"])
+    res = client.get("/index/?baseid=baseid_123")
```
paulineribeyre: cool, please use `doc["base_id"]` instead of `baseid_123` here, so that the test doesn't break if someones change the data later

tests/test_client.py
```diff
@@ -27,312 +26,469 @@ def get_doc(
...
+
+    res = client.get("/index/?uploader=uploader_123")
+    assert res.status_code == 200
+    rec = res.json
+    assert rec["records"][0]["uploader"] == "uploader_123"
+    assert not rec["records"][0]["file_name"]
 
-    r = swg_index_client.list_entries(uploader="uploader_123")
-    assert r.records[0].uploader == "uploader_123"
-    assert not r.records[0].file_name
```
paulineribeyre: these 2 lines should not have been deleted

tests/test_client.py
```diff
@@ -27,312 +26,469 @@ def get_doc(
...
+
+    res = client.get("/index/?uploader=uploader_321")
+    assert res.status_code == 200
+    rec = res.json
+    assert rec["records"][0]["uploader"] == "uploader_321"
+    assert rec["records"][0]["file_name"] == "myfile.txt"
 
-    r = swg_index_client.list_entries(uploader="uploader_321")
-    assert r.records[0].uploader == "uploader_321"
-    assert r.records[0].file_name == "myfile.txt"
```
paulineribeyre: these 2 lines should not have been deleted

tests/util.py
```diff
@@ -34,13 +34,11 @@ def assert_blank(r):
...
-    assert not r.records[0].hashes.md5
-    assert not r.records[0].hashes.sha
-    assert not r.records[0].hashes.sha256
-    assert not r.records[0].hashes.sha512
+    assert r["records"][0]["baseid"]
+    assert r["records"][0]["did"]
+    assert not r["records"][0]["size"]
+    assert not r["records"][0]["acl"]
+    assert not r["records"][0]["authz"]
+    assert not r["records"][0]["hashes"]
```
paulineribeyre: the blank record used to have this (master branch):
```
'hashes': {'crc': None,
            'etag': None,
            'md5': None,
            'sha': None,
            'sha256': None,
            'sha512': None}
```
and now it has `'hashes': {}`
I think maybe because of the way you GET it? can you check what's going on?

tests/test_client.py
callenmacphee: Are these two not mirrored by the two lines I added above?

paulineribeyre: ha, i missed that, sorry :p

tests/util.py
callenmacphee: I changed this because the response of the GET leads to a key error when the keys within the hashes dictionary don't exist and I thought that the individual key assertions would be subsumed by the empty hash assertion, although I may be wrong about this and could find another work around if so.

tests/test_client.py
callenmacphee: no worries!

tests/util.py
paulineribeyre: but why did the response contain all the keys before and not now?

paulineribeyre: Note on this: looks like the swagger client was creating default "None" values for each hash type

PR #235, opened 2019-08-12
test(leftovers): add tests to hit uncovered code

callenmacphee:
Adds miscellaneous tests to cover leftover uncovered code. 

### Improvements
Adds missing unit tests


tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    rec = res.json
+    response = get_record(rec["did"])
+    record = response[0].json
+    assert record["urls"][0] == "s3://endpointurl/bucket/key"
+    assert record["size"] == 123
+
+    # test call dist_get_record for unposted record
+    fake_did = "testprefix:d96bab16-c4e1-44ac-923a-04328b6fe78f"
+    with pytest.raises(IndexNoRecordFound):
+        get_record(fake_did)
```
paulineribeyre: also use the `GET /<GUID>` endpoint here, and i think you'll have to check the status code instead of looking for `IndexNoRecordFound`

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    rec = res.json
+    response = get_record(rec["did"])
+    record = response[0].json
+    assert record["urls"][0] == "s3://endpointurl/bucket/key"
+    assert record["size"] == 123
+
+    # test call dist_get_record for unposted record
```
paulineribeyre:
```suggestion
    # test getting an ID that does not exist
```

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
+from indexd.blueprint import get_record
+from tests.test_client import get_doc
+import pytest
+from indexd.index.errors import NoRecordFound as IndexNoRecordFound
+from indexd.dos.blueprint import get_dos_record
+import json
+
+
+def test_get_record_call_dist_get_record(client, user):
+    # test get_record correctly
```
paulineribeyre:
```suggestion
    # test getting an existing ID
```

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+def test_get_dos_record(client, user):
+    data = get_doc(has_urls_metadata=True, has_metadata=True, has_baseid=True)
+
+    res_1 = client.post("/index/", json=data, headers=user)
+    assert res_1.status_code == 200
+
+    # test exception raised at nonexistent
+    fake_did = "16e51a9a-af23-45a3-b4b2-94ae2b143feb"
+    with pytest.raises(IndexNoRecordFound):
+        get_dos_record(fake_did)
```
paulineribeyre: use the endpoint instead

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+        "host_authorities": ["PDC"],
+    }
+    ark = "ark:/31807/TEST-abc"
+    res = client.put("/alias/" + ark, json=data, headers=user)
+    assert res.status_code == 200
+    rec = res.json
+    assert rec["name"] == ark
+    # assert there is only one entry in the alias index
+    assert len(client.get("/alias/").json["aliases"]) == 1
+    res = client.get("/alias/?hash=" + "md5:8b9942cf415384b27cadf1f4d2d682e5")
```
paulineribeyre:
```suggestion
    res = client.get("/alias/?hash=" + data["hashes"]["md5"])
```

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    }
+    ark = "ark:/31807/TEST-abc"
+    res = client.put("/alias/" + ark, json=data, headers=user)
+    assert res.status_code == 200
+    rec = res.json
+    assert rec["name"] == ark
+    # assert there is only one entry in the alias index
+    assert len(client.get("/alias/").json["aliases"]) == 1
+    # assert that the returned alias by size is the same as the posted
+    res = client.get("/alias/?size=" + str(data["size"]))
```
paulineribeyre: using `format` to build strings lets you use ints without having to convert to strings :D

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+        "keeper_authority": "CRI",
+        "host_authorities": ["PDC"],
+    }
+    ark = "ark:/31807/TEST-abc"
+    res = client.put("/alias/" + ark, json=data, headers=user)
+    assert res.status_code == 200
+    rec = res.json
+    assert rec["name"] == ark
+    # assert there is only one entry in the alias index
+    assert len(client.get("/alias/").json["aliases"]) == 1
```
paulineribeyre: this is a bit dangerous; what if the `GET` result is not 200? then there will be no `json` field and you'll get an error
it would be safer to separate this into multiple steps, `GET` - check status code - get json - check that there's only one alias

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
```
paulineribeyre: same comment here

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
+from indexd.blueprint import get_record
+from tests.test_client import get_doc
+import pytest
+from indexd.index.errors import NoRecordFound as IndexNoRecordFound
+from indexd.dos.blueprint import get_dos_record
+import json
+
+
+def test_get_record_call_dist_get_record(client, user):
```
paulineribeyre: i think this test should be renamed to something that makes it easier to understand what is being tested

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    data3["size"] = 300
+    res = client.post("/index/", json=data3, headers=user)
+    assert res.status_code == 200
+    data_size = data1["size"] + data2["size"] + data3["size"]
+    index_stats = client.get("/_stats/").json
+    file_number = index_stats["fileCount"]
+    index_file_size = index_stats["totalFileSize"]
+    # test that the stat file number and size is consistent with post
+    assert file_number == 3
+    assert index_file_size == data_size
```
paulineribeyre: cool test!!

you could make the assertions more compact by using directly `assert index_stats["fileCount"] == 3` and `assert index_stats["totalFileSize"] == data_size`

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def test_status_check(client, user):
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
```
paulineribeyre: why is this part of the health check? :)

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    assert res.status_code == 200
+    res = client.get("/_status/")
+    assert res.status_code == 200
+
+
+def test_version_check(client, user):
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
```
paulineribeyre: why is this part of the version check?

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+
+def test_index_list_by_file_name(client, user):
+    # post two records of different name
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    data["file_name"] = "test_file"
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    res = client.get("/index/?file_name=" + data["file_name"])
```
paulineribeyre: your test case is `1 record with a file name + 1 record with no file name; query with the file name`. but this doesn't test that records that _have_ file names that are different from the one you query are not returned

could you add a 3rd record with a different file name?

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+
+def test_index_list_by_authz(client, user):
+    # post two records of different authz
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    data["authz"] = ["test_authz"]
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    res = client.get("/index/?authz=" + data["authz"][0])
```
paulineribeyre: same comment

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+
+def test_index_list_by_version(client, user):
+    # post two records of different version
+    data = get_doc()
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    data["version"] = "3"
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
+    res = client.get("/index/?version=" + data["version"])
```
paulineribeyre: same comment :p

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    res_1 = client.post("/index/", json=data1, headers=user)
+    assert res_1.status_code == 200
+
+    data2 = get_doc()
+    data2["file_name"] = "test_file_name_2"
+    res_2 = client.post("/index/", json=data2, headers=user)
+    assert res_2.status_code == 200
+
+    negate_param = {"file_name": "test_file_name_2"}
+    res = client.get("/index/?negate_params=" + json.dumps(negate_param))
```
paulineribeyre: sort of the same comment here; could you add a 3rd record that has _no_ file name? from the docs, i think this endpoint should also return records that don't have the "negate param"

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    data1["file_name"] = "test_file_name_1"
+    res_1 = client.post("/index/", json=data1, headers=user)
+    assert res_1.status_code == 200
+
+    data2 = get_doc()
+    data2["file_name"] = "test_file_name_2"
+    res_2 = client.post("/index/", json=data2, headers=user)
+    assert res_2.status_code == 200
+
+    negate_param = {"file_name": "test_file_name_2"}
```
paulineribeyre:
```suggestion
    negate_param = {"file_name": data2["file_name"]}
```

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    res = client.get("/index/?negate_params=" + json.dumps(negate_param))
+    assert res.status_code == 200
+    rec = res.json
+    # assert record returned with proper non-negated file name
+    assert len(rec["records"]) == 1
+    assert rec["records"][0]["file_name"] == data1["file_name"]
+
+
+def test_negate_filter_acl(client, user):
```
paulineribeyre: i would argue that you don't need these 4 `test_negate_filter_file_name`, `test_negate_filter_acl`, `test_negate_filter_authz` and `test_negate_filter_version`. there are 3 different cases for this feature: `negate_filter` can be a string, a list or a dict. for dict, there are 2 cases (see details [here](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/uc-cdis/Indexd/master/openapis/swagger.yaml#/index/listEntries))

imo you should keep `test_negate_filter_file_name` and `test_negate_filter_authz`, remove the other 2 (unless they are adding coverage -> lmk if i'm wrong) and add a test for either `metadata` or `urls_metadata` that includes both dict use cases

tests/test_leftovers.py
```diff
@@ -0,0 +1,294 @@
...
+    data = {"size": 123, "hashes": {"md5": "8b9942cf415384b27cadf1f4d2d682e5"}}
+    fake_did = "testprefix:455ffb35-1b0e-49bd-a4ab-3afe9f3aece9"
+    fake_rev = "8d19b5c10"
+    res = client.put(
+        "/index/blank/{}?rev={}".format(fake_did, fake_rev), json=data, headers=user
+    )
+    assert res.status_code == 404
+
+
+def test_authz_response(client, user):
```
paulineribeyre:
```suggestion
def test_unauthorized_create(client):
```

tests/test_leftovers.py
callenmacphee: these do add to the test coverage, but I can definitely remove them if they don't add any actual value.

tests/test_leftovers.py
callenmacphee: I believe it returns either the negate param or the blank keys for that param

tests/test_leftovers.py
callenmacphee: I modified this to having md5 inside the first string, I can change it to be more variable but it seemed maybe a little messy and it seems as though we use md5 hashing for the most part?

tests/test_leftovers.py
paulineribeyre: then let's keep them :)

tests/test_leftovers.py
```diff
@@ -312,7 +310,7 @@ def test_negate_filter_version(client, user):
     assert res_2.status_code == 200
 
     data3 = get_doc()
-    res_2 = client.post("/index/", json=data2, headers=user)
+    res_2 = client.post("/index/", json=data3, headers=user)
     assert res_2.status_code == 200
```
paulineribeyre: it works either way but this should probably be `res3` for consistency

tests/test_leftovers.py
```diff
@@ -1,38 +1,39 @@
...
+    res = client.post("/index/", json=data, headers=user)
+    assert res.status_code == 200
 
     # test exception raised at nonexistent
-    fake_did = "16e51a9a-af23-45a3-b4b2-94ae2b143feb"
-    with pytest.raises(IndexNoRecordFound):
-        get_dos_record(fake_did)
+    fake_did = "testprefix:d96bab16-c4e1-44ac-923a-04328b6fe78f"
+    res = client.get("/ga4gh/dos/v1/dataobjects/" + fake_did)
+    assert res.status_code == 500
```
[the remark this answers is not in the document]
callenmacphee: I forgot to mention in the last comments my bad, I'm unable to get a 404 response from this and I was wondering if you would know why? The endpoint returns a 200 when I switch out the fake_did for a real posted one, so I thought it was formatted correctly. Is there another way to get a 404 besides what I'm doing?

tests/test_client.py
```diff
@@ -98,6 +98,71 @@ def test_index_list_with_params(client, user):
     assert data_list["records"][0]["did"] == rec_1["did"]
     assert data_list["records"][0]["urls_metadata"] == data1["urls_metadata"]
 
+    # post two records of different size
```
paulineribeyre: could you separate these back into `test_index_list_by_size`, `test_index_list_by_file_name`, etc? i know this test could include everything but it's just way too big @_@

tests/test_client.py
```diff
@@ -199,6 +264,99 @@ def test_index_list_with_params_negate(client, user):
     assert rec_4["did"] in ids
     assert rec_5["did"] in ids
 
+    # post two records of different file name
```
paulineribeyre: same here :/ sorry lol

tests/test_leftovers.py
paulineribeyre: i didn't see the dict use case though

tests/test_leftovers.py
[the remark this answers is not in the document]
paulineribeyre: i'll have a look tomorrow

tests/test_client.py
callenmacphee: no worries, I agreee :7

tests/test_leftovers.py
```diff
@@ -1,38 +1,39 @@
...
+    # test getting an existing ID
     data = get_doc()
     res = client.post("/index/", json=data, headers=user)
     assert res.status_code == 200
     rec = res.json
-    response = get_record(rec["did"])
-    record = response[0].json
-    assert record["urls"][0] == "s3://endpointurl/bucket/key"
-    assert record["size"] == 123
+    response = client.get("/index/" + rec["did"])
```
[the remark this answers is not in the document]
callenmacphee: good catch, I think it was covered very sneakily in one assert statement of the original tests. I wrote another more obvious test for it as well just in case, I can delete it though if need be.

tests/test_leftovers.py
paulineribeyre: note on this: it's in `test_index_list_with_params_negate`

tests/test_client.py
```diff
@@ -1256,6 +1640,18 @@ def test_dos_get(client, user):
     assert rec_3["data_object"]["id"] == rec_1["did"]
 
 
+def test_get_dos_record_error(client, user):
+    data = get_doc(has_urls_metadata=True, has_metadata=True, has_baseid=True)
+
+    res = client.post("/index/", json=data, headers=user)
```
paulineribeyre: this record you created is not used later in the test. you can remove it

callenmacphee: cool got it!

PR #248, opened 2019-10-21
Feat/index record alias api

mpingram:
PXP-4353
Implement API to manipulate index_record_alias

* Add ability to create, update and delete aliases for GUIDs. Aliases can be used to query the root endpoint for a record instead of that record's GUID. Aliases are associated with one GUID only.

* API uses `create` and `update` Arborist permissions

* Change IndexRecordAlias table -- add unique constraint to alias name column

### New Features
- Implement API to manipulate aliases for Indexd records.

### Bug Fixes
- Fix bug in Indexd DOS search (GET `/ga4gh/dos/v1/dataobjects/<path:record>`) preventing aliases from being returned when querying for a record's baseID.


indexd/index/blueprint.py
```diff
@@ -374,6 +375,74 @@ def add_index_record_version(record):
...
+    Get all aliases associated with this DID / GUID
+    """
+    try:
+        aliases = blueprint.index_driver.get_aliases_for_did(record)
+        aliases_payload = {
+            "aliases": [{"value": alias} for alias in aliases]
+        }
+        return flask.jsonify(aliases_payload), 200
+    except NoRecordFound as err:
+        raise err
```
Avantol13: we could add some logs here with more details including the `record` that got passed in (same for errors below)

Avantol13: can use our `cdislogging` library. something like: 

```
from cdislogging import get_logger

from gen3authz.client.arborist.client import ArboristClient
logger = get_logger(__name__, log_level="info")  # TODO that info level should probably be set based on some configuration to "debug" but unsure if indexd has such a thing right now
```

and then in the funcitons:
```
def do_something(foobar):
  logger.info("info!!")
  logger.warning("warning!!")
  logger.error(f"very bad things have occurred. error: {err}")
```

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+    assert res.status_code == 200
+    record = res.get_json()
+    assert record["did"] == guid_B
+    assert record["did"] != guid_A
+
+def test_global_endpoint_alias_endpoint_collision(client, guid, aliases, user):
+    """
+    when an alias shares a name with an endpoint on the root of
+    the API, querying that endpoint should function as normal, 
+    not resolve the alias
```
paulineribeyre: nice!

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+    """
+    # associate a new alias named "index/" with guid. "index/" is an endpoint
+    # on the root of the api.
+    aliases_payload = to_payload(["index"])
+    res = client.post(get_endpoint(guid), json=aliases_payload, headers=user)
+    assert res.status_code == 200
+
+    # when we query "_status", we expect to see the "_status" endpoint, not
```
paulineribeyre:
```suggestion
    # when we query "index", we expect to see the "index" endpoint, not
```

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+    # an alias.
+    res = client.get("index")
+    assert res.status_code == 200
+    body = res.get_json()
+    body_is_not_record = body.get("did", None) is None
```
paulineribeyre: `None` already is `get`'s default default (💥)

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+    expected_aliases = aliases
+    assert set(aliases_in_db) == set(expected_aliases)
+
+def test_POST_aliases_GUID_already_has_alias(client, user, guid, aliases):
+    """
+    expect to return 400 and have no effect if valid GUID and one or more aliases 
+    already associated with this GUID
+    """
+    # pick a random subset of the aliases already associated with this GUID
+    subset_old_aliases = aliases[0:1]
```
paulineribeyre: about the comment: that's not random :p but it's fine haha

indexd/index/blueprint.py
```diff
@@ -374,6 +375,74 @@ def add_index_record_version(record):
...
+def append_aliases(record):
+    """
+    Append one or more aliases to aliases already associated with this
+    DID / GUID, if any.
+    """
+    aliases_json = flask.request.get_json()
+    try:
+        jsonschema.validate(aliases_json, RECORD_ALIAS_SCHEMA)
+    except jsonschema.ValidationError as err:
+        raise UserError(err)
```
paulineribeyre: what does this error look like? does it clearly state that the body doesn't have the expected format?

paulineribeyre:
please format my code

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+def aliases(client, user, guid):
+    """
+    Associates aliases with a GUID in indexd and returns the new aliases.
+    """
+    aliases = [
+        '.?=G}k@Up3LIlv+p96yaI06,t@?j=ejk[%+', 
+        'Fa"uW<A"/\'hELmTjH%r%6@Tp^HaB^', 
+        '{j\'8D6d5fc]5#[*9n%|G9"hZ?z3:wX', 
+        '"U7XsT+EXD|1?@$ywDV"ce<B}7v9t)'
+    ]
```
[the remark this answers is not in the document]
paulineribeyre: how many characters does the db allow? what happens if the name is longer?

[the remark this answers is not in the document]
mpingram:
```
mpingram=> \d index_record_alias
   Table "public.index_record_alias"
 Column |       Type        | Modifiers
--------+-------------------+-----------
 did    | character varying | not null
 name   | character varying | not null
```
Doesn't seem like there is a db limit to number of characters in an alias. (`If character varying is used without length specifier, the type accepts strings of any size.` from https://www.postgresql.org/docs/11/datatype-character.html). Should we enforce one?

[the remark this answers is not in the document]
paulineribeyre: no that's great!

indexd/index/blueprint.py
mpingram: The jsonschema error looks like this: https://python-jsonschema.readthedocs.io/en/stable/errors/#jsonschema.exceptions.ValidationError.message
The jsonschema error should have all the info required, unless we want to wrap it in an error of our own?

paulineribeyre: if it contains enough information for the user to fix the input, it's all good

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+# POST /index/{GUID}/aliases
+# -------------------------
+def test_POST_aliases_valid_GUID_valid_aliases(client, user, guid, aliases, unused_aliases):
+    """
+    normal operation: expect to append to aliases and return list of all aliases
+    """
+    new_aliases = unused_aliases
+    new_alias_payload = to_payload(new_aliases)
+    res = client.post(get_endpoint(guid), json=new_alias_payload, headers=user)
+    assert res.status_code == 200
```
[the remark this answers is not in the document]
mpingram: Nice catch, tyvm

tests/test_aliases_endpoints.py
```diff
@@ -0,0 +1,557 @@
...
+    guid = res.get_json()["did"]
+
+    return guid
+
+@pytest.fixture(scope="function")
+def guid(client, user):
+    """
+    Creates a random record in indexd and returns that record's GUID.
+    """
+    return create_random_record(client, user)
```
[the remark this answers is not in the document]
mpingram: My reasoning was `create_random_record` is called in the test code too!

mpingram:
> please check whether the README's mentions of aliases shoulb be updated

README looks good to me, it describes aliases as linking to GUIDs (which is the case with the changes from this PR)

mpingram:
wool

tests/test_aliases_endpoints.py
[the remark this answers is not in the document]
paulineribeyre: ah right!

indexd/index/blueprint.py
```diff
@@ -432,9 +434,11 @@ def replace_aliases(record):
     aliases = [record["value"] for record in aliases_json["aliases"]]
 
     # authorization and error handling done in driver
-    ret = blueprint.index_driver.replace_aliases_for_did(aliases, record)
+    blueprint.index_driver.replace_aliases_for_did(aliases, record)
 
-    return flask.jsonify(ret), 200
+    aliases = blueprint.index_driver.get_aliases_for_did(record)
```
paulineribeyre: instead of querying the db, `replace_aliases_for_did` can return the list

PR #251, opened 2019-11-07
PXP-1089 add test + fix acl/authz/size query

paulineribeyre:
### Bug Fixes
- Fix query by multiple ACL items or multiple authz items or multiple URLs
- Fix query by URLs with size=0

### Improvements
- Improve the OpenAPI documentation with more details on how to query by multiple items of the same parameter


indexd/index/drivers/alchemy.py
```diff
@@ -374,17 +374,31 @@ def ids(
                 for u in urls:
                     query = query.filter(IndexRecordUrl.url == u)
 
+            # filter records that have ALL the ACL elements
             if acl:
-                query = query.join(IndexRecord.acl)
+                sub = session.query(IndexRecordACE.did)
```
vpsx: Would we lose anything by doing this instead?
```
if acl:
    for u in acl:
        sub = session.query(IndexRecordACE.did).filter(IndexRecordACE.ace == u)
        query = query.filter(IndexRecord.did.in_(sub.subquery()))
```

indexd/index/drivers/alchemy.py
```diff
@@ -374,17 +374,31 @@ def ids(
...
+                    else:
+                        dids = dids.intersection(e[0] for e in subsub.all())
+                query = query.filter(IndexRecord.did.in_(dids))
             elif acl == []:
                 query = query.filter(IndexRecord.acl == None)
 
+            # filter records that have ALL the authz elements
             if authz:
-                query = query.join(IndexRecord.authz)
+                sub = session.query(IndexRecordAuthz.did)
```
vpsx: as with acl

paulineribeyre:
> do we have any particular reason for assuming that `?x=y,z` means AND instead of OR?

if users want OR they can query each item separately; this is the only way to query AND
(note that currently `?x=y,z` doesn't return anything)

> do we want the same behavior for urls?

mm, i guess 🤔

PR #253, opened 2019-11-21
HSClient handling

paulineribeyre:
dataguids.org uses the feat/hs branch - merge it into master so that we can pin dataguids.org

### New Features
- Handle HSClient to interact with the Commons Share Object API


indexd/dos/blueprint.py
```diff
@@ -120,22 +127,26 @@ def indexd_to_dos(record):
 
 @blueprint.errorhandler(UserError)
 def handle_user_error(err):
-    return flask.jsonify(error=str(err)), 400
+    ret = {"msg": str(err), "status_code": 400}
```
jiaqi216: why do you need to return the status code also in the response data?

paulineribeyre: i don't know why it was done this way, but a production website is using a branch with this change so i'd rather not change the contents of the response

PR #254, opened 2019-11-24
feat(page): add page param to support pagination without knowing dids

Avantol13:
### New Features
- add `page` param to list records endpoint to support pagination without using a did

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -323,6 +323,7 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
     def ids(
         self,
         limit=100,
+        page=None,
         start=None,
         size=None,
         urls=None,
@@ -439,6 +440,9 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
             else:
                 # only apply limit when ids is not provided
                 query = query.limit(limit)
+
+            if page is not None:
+                query = query.offset(limit * page)
 
             return [i.to_document_dict() for i in query]
 
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -433,7 +433,16 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
             if urls_metadata or negate_params:
                 query = query.distinct(IndexRecord.did)
 
-            query = query.order_by(IndexRecord.did)
+            if page:
+                # when we're paginating, order by updated date so newly added stuff is
+                # at the end (reduce risk that a new records ends up in a page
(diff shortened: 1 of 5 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -432,14 +433,26 @@ def ids(
...
+            if page is not None:
+                # when we're paginating, order by updated date so newly added stuff is
+                # at the end (reduce risk that a new records ends up in a page
+                # earlier on) and allows for some logic to check for newly added records
+                # (e.g. parallelly processing from beginning -> middle and ending -> middle
+                #       and as a final step, checking the "ending"+1 to see if there are
+                #       new records).
+                query = query.order_by(IndexRecord.updated_date)
+            else:
+                query = query.order_by(IndexRecord.did)
```
paulineribeyre: should we just always order by date?

indexd/index/drivers/alchemy.py
```diff
@@ -432,14 +433,26 @@ def ids(
...
 
-            query = query.order_by(IndexRecord.did)
+            if page is not None:
+                # when we're paginating, order by updated date so newly added stuff is
+                # at the end (reduce risk that a new records ends up in a page
+                # earlier on) and allows for some logic to check for newly added records
+                # (e.g. parallelly processing from beginning -> middle and ending -> middle
+                #       and as a final step, checking the "ending"+1 to see if there are
+                #       new records).
+                query = query.order_by(IndexRecord.updated_date)
```
paulineribeyre: why not order by decreasing date? if we use this to validate after indexing for example, we could just get the first pages

indexd/index/drivers/alchemy.py
Avantol13: I thought about that... probably? I was afraid to change the existing behavior tbh, but I guess it shouldn't matter, you'll always get the next `did` regardless of the ordering

indexd/index/drivers/alchemy.py
Avantol13: The way this is set the dates later in the past will be first such that any additional records end up at the end. This means that existing pages won't change content, whereas if we were to do the opposite, a new record would bump everything else forward, changing the content on all existing pages. So I didn't want to break things mid-processing. If you're curious, this is the logic I was cross-testing with gen3 sdk: https://github.com/uc-cdis/gen3sdk-python/blob/907d418ebcf46ea345472229982c604c54cce179/gen3/tools/indexing.py#L82

paulineribeyre: makes sense!! thanks

PR #261, opened 2020-02-27
Drs endpoints

BinamB:
Description about what this pull request does.

Please make sure to follow the [DEV guidelines](https://gen3.org/resources/developer/dev-introduction/) before asking for review.

### New Features
- Implemented new DRS endpoints that follows paths a parameters that follows the DRS spec https://ga4gh.github.io/data-repository-service-schemas/docs/#_paths
- Update swagger docs to include DRS
- Added unit tests for DRS endpoints


indexd/drs/blueprint.py
```diff
@@ -0,0 +1,153 @@
...
+def get_drs_object(object_id):
+    """
+    Returns a specific DRSobject with object_id
+    """
+    ret = blueprint.index_driver.get(object_id)
+    return flask.jsonify(indexd_to_drs(ret)), 200
+
+    
+@blueprint.route("/ga4gh/drs/v1/objects", methods=["GET"])
+def list_drs_records():
```
paulineribeyre: TODO see if we can just call the existing list function from "/index"
and call `indexd_to_drs` on each record

indexd/drs/blueprint.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # access_methods mapping
+    if "urls" in record:
+        drs_object["access_methods"] = []
+        for location in record["urls"]:
+            location_type = location.split(":")[0] #(s3, gs, ftp, gsiftp, globus, htsget, https, file)
+
+            drs_object["access_methods"].append({
+                "type":location_type, 
+                "access_url":get_signed_url_for_object(record["did"], "") if bearer_token and not list_drs else {"url":location},
```
paulineribeyre: do you not need `access_id` in this call to `get_signed_url_for_object`? if `access_id` is optional, i'd suggest using a default value in the function definition of `get_signed_url_for_object` instead of passing an empty string here

indexd/fence_client.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, object_id, access_id):
+        self.object_id = object_id
+        self.access_id = access_id
+
+    @cached_property
+    def get_signed_url_from_fence(self):
+        fence_server = (flask.current_app.config.get("PRESIGNED_URL_ENDPT")
+                            or flask.current_app.presigned_url_endpt)
+        api_url = fence_server.rstrip("/") + "/data/download/"
+        url = api_url + self.object_id + "?protocol=" + self.access_id
```
paulineribeyre: in the `indexd_to_drs` function you use an empty string as `access_id`, so maybe do this? it probably works either way but the URL would be cleaner
```
if self.access_id:
    # add protocol param
```

indexd/drs/blueprint.py
```diff
@@ -0,0 +1,153 @@
...
+def get_signed_url(object_id, access_id=""):
+    res = get_signed_url_for_object(object_id, access_id)
+    if not res:
+        raise IndexNoRecordFound("No signed url found")
+    return res, 200
+
+
+def indexd_to_drs(record, list_drs = False):
+    bearer_token = flask.request.headers.get('AUTHORIZATION')
+    drs_object = {"id": record["did"], "description": "", "mime_type": ""}
```
paulineribeyre: - file_name, created_date, updated_date, size, urls: these fields always exist in records so you don't need the `if` statements
- mime_type  should be `application/json`
- self_uri doesn't exist in indexd records, and the DRS spec says it's required. i think this current code would never include it, maybe @jiaqi216 can help figuring out what value to use
- contents and alias don't exist in indexd records, they are optional in the spec, they should probable always be included, but empty for now

indexd/fence_client.py
```diff
@@ -0,0 +1,53 @@
...
+class FenceRequest(object):
+
+    def __init__(self, object_id, access_id):
+        self.object_id = object_id
+        self.access_id = access_id
+
+    @cached_property
+    def get_signed_url_from_fence(self):
+        fence_server = (flask.current_app.config.get("PRESIGNED_URL_ENDPT")
+                            or flask.current_app.presigned_url_endpt)
```
paulineribeyre:
```suggestion
        fence_server = flask.current_app.config.get("PRESIGNED_URL_ENDPT", "http://.....")
```

indexd/default_settings.py
```diff
@@ -32,6 +32,8 @@
     )
 }
 
+CONFIG["PRESIGNED_URL_ENDPT"] = "http://presigned-url-fence-service/"
```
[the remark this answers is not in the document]
BinamB: I did this because `presigned-url-fence-service` has been separated from `fence-service` so `FENCE_URL` , which would be `http://fence-service`, is different then the `http://presigned-url-fence-service/' which contains the "/data/download". Should I still changed it to `FENCE_URL`?

indexd/drs/blueprint.py
BinamB: Doing this because the endpoint `/ga4gh/drs/v1/objects/<path:object_id>` when given a bearer token should have presigned_url in the url field. Since we don't need an access_id for this endpoint but access_id is **required** for `/ga4gh/drs/v1/objects/<path:object_id>/access/<path:access_id>`

openapis/swagger.yaml
```diff
@@ -1603,6 +1751,205 @@ definitions:
...
+      msg:
+        type: string
+        description: A detailed error message.
+      status_code:
+        type: integer
+        description: The integer representing the HTTP status code (e.g. 200, 404).  
+  ListDrsObject:
+    type: object
+    properties:
+      DrsObjects:
```
paulineribeyre:
```suggestion
      drs_objects: 
```

openapis/swagger.yaml
```diff
@@ -894,6 +896,151 @@ paths:
           type: string
       tags:
         - DOS
+  '/ga4gh/drs/v1/objects':
+    get:
+      summary: List all DrsObject
+      description: ''
```
paulineribeyre: i think you should mention whether URLs are signed or not

openapis/swagger.yaml
```diff
@@ -894,6 +896,151 @@ paths:
...
+      operationId: GetObject
+      responses:
+        '200':
+          description: The DrsObject was found successfully.
+          schema:
+            $ref: '#/definitions/DrsObject'
+        '202':
+          description: >
+            The operation is delayed and will continue asynchronously.
+            The client should retry this same request after the delay specified by Retry-After header.
```
paulineribeyre: do we really return 202??

openapis/swagger.yaml
```diff
@@ -894,6 +896,151 @@ paths:
...
+        - name: object_id
+          in: path
+          required: true
+          type: string
+        - in: query
+          name: expand
+          type: boolean
+          default: false
+          description: >-
+            If false and the object_id refers to a bundle, then the ContentsObject array
```
paulineribeyre:
```suggestion
          description: >-
            NOT IMPLEMENTED YET.
            If false and the object_id refers to a bundle, then the ContentsObject array
```

openapis/swagger.yaml
```diff
@@ -894,6 +896,151 @@ paths:
...
+          schema:
+            $ref: '#/definitions/ListDrsObject'
+      security: []
+      tags:
+        - DRS
+  '/ga4gh/drs/v1/objects/{object_id}':
+    get:
+      summary: Get info about a DrsObject.
+      description: >-
+        Returns object metadata, and a list of access methods that can be used to fetch object bytes.
```
paulineribeyre: i think you should mention in which conditions URLs are signed

indexd/drs/blueprint.py
```diff
@@ -0,0 +1,153 @@
...
+import requests
+
+blueprint = flask.Blueprint("drs", __name__)
+
+blueprint.config = dict()
+blueprint.index_driver = None
+blueprint.dist = []
+
+@blueprint.route("/ga4gh/drs/v1/objects/<path:object_id>", methods=["GET"])
+def get_drs_object(object_id):
```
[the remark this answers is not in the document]
paulineribeyre: i'm still unsure about this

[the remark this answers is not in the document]
BinamB: Is this what you meant? When given an access token, url is the presigned url but when not given then url is the location. Have tested for both the cases:
https://github.com/uc-cdis/indexd/blob/drs-endpoints/tests/test_drs.py#L82

indexd/fence_client.py
```diff
@@ -28,7 +30,11 @@ def get_signed_url_for_object(self, object_id, access_id):
         fence_server = self.url
         api_url = fence_server.rstrip("/") + "/data/download/"
         url = api_url + object_id
-        if access_id:
+        if access_id is not None and access_id != "":
+            if access_id not in SUPPORTED_PROTOCOLS:
+                raise UserError(
+                    "The specified protocol {} is not supported".format(access_id)
+                )
```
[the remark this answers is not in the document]
BinamB: The thing with this is that, given the incorrect protocol, fence doesn't return an error that indexd can read as far as I know. If I can remember correctly it either gives me some kind of 404 instead of a 400.
ex: _{"msg":"No document with id:blah1234 with access_id:s2","status_code":404}_ which i thought was misleading. I dont know. What do you suggest?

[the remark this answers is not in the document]
paulineribeyre: i think it's fine to return an error that lets the user know either the guid is not found or the access_id is not found. looking at the fence code, it looks like it returns a 400

tests/test_drs.py
```diff
@@ -0,0 +1,192 @@
...
+        "https://fictitious-commons.io/data/download/" + did + "?protocol=" + protocol
+    )
+    presigned_url = {
+        "url": "https://storage.googleapis.com/nih-mock-project-released-phs123-c2/RootStudyConsentSet_phs000007.Whatever.v666.p1.c2.FBI-BMW-CIA.tar.gz?GoogleAccessId=internal-someuser-1399@dcpstage-210518.iam.gserviceaccount.com&Expires=1582215120&Signature=hUsgjkegdsfkjbsajkafnsdjksdnfjknbdsajkfbsdkjfbjdfbkjdasfbnjsdnfjsnd2FTr%2FKs2kGKs0fJ8v5elFk5NQAYdrGcU3kROrzJuHUbI%2BMZ839SAbAz2rbMBuC9e46%2BdB91%2FA==&userProject=dcf-mock-project"
+    }
+    responses.add(responses.GET, full_url, json=presigned_url, status=status)
+    return presigned_url
+
+
+def get_doc(has_version=True, multiple_endpointurl=False, has_content=False):
```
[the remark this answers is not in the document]
BinamB: Its for when we implement bundles. should I still keep it?

tests/test_drs.py
```diff
@@ -0,0 +1,192 @@
...
+    res_2 = client.get("/ga4gh/drs/v1/objects/" + rec_1["did"])
+
+    assert res_2.status_code == 200
+    rec_2 = res_2.json
+    assert rec_2["id"] == rec_1["did"]
+
+    endpointurls = [
+        "sftp://endpointurl/bucket/key",
+        "ftp://endpointurl/bucket/key",
+        "gs://endpointurl/bucket/key",
```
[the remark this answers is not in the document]
BinamB: You genius!

tests/test_drs.py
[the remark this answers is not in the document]
paulineribeyre: i'd remove it until we implement bundles 🤷‍♀

tests/test_drs.py
```diff
@@ -0,0 +1,176 @@
...
+        "form": "object",
+        "size": 123,
+        "urls": ["s3://endpointurl/bucket/key"],
+        "hashes": {"md5": "8b9942cf415384b27cadf1f4d2d682e5"},
+    }
+    if has_version:
+        doc["version"] = "1"
+    if multiple_endpointurl:
+        for url_type in ["gs", "ftp", "sftp"]:
+            doc["urls"].append("{}://endpointurl/bucket/key".format(url_type))
```
paulineribeyre: the `urls` param is not used right now, it should be used to set the urls

my suggestion was to get rid of the `multiple_endpointurl` param so that you don't have to hardcode `://endpointurl/bucket/key` in this function, which lets the tests control which urls to use, and check the result more easily

tests/test_drs.py
```diff
@@ -0,0 +1,176 @@
...
+    assert rec_2["id"] == rec_1["did"]
+    assert rec_2["size"] == data["size"]
+    for k in data["hashes"]:
+        assert rec_2["checksums"][0]["checksum"] == data["hashes"][k]
+        assert rec_2["checksums"][0]["type"] == k
+
+    assert rec_2["access_methods"][0]["access_url"] == presigned
+
+
+def test_drs_list(client, user):
```
paulineribeyre: this test is fine but it doesn't really test the listing functionality since there is only 1 record

PR #264, opened 2020-03-20
WIP feat(redis): redis client and caching logic

Avantol13:
:warning: Work in progress :warning: 

- Describe what this pull request does.
- Add short descriptive bullet points for each section if relevant. Keep in mind that they will be parsed automatically to generate official release notes.
- Test manually.
- Maintain or increase the test coverage (if relevant).
- Update the documentation, or justify if not needed.

-->

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -223,17 +223,16 @@ def get_index_record(record):
     # check redis
     json_record = flask.current_app.redis_client.get(record)
     if json_record:
-        json_record = json_record.loads(json_record)
+        json_record = json.loads(json_record)
     else:
         # get from db
-        ret = blueprint.index_driver.get(record)
-        json_record = flask.jsonify(ret)
+        json_record = blueprint.index_driver.get(record)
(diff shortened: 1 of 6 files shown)


indexd/index/blueprint.py
```diff
@@ -220,9 +220,19 @@ def get_index_record(record):
     Returns a record.
     """
 
-    ret = blueprint.index_driver.get(record)
+    # check redis
+    json_record = flask.current_app.redis_client.get(record)
+    if json_record:
+        json_record = json.loads(json_record)
+    else:
+        # get from db
```
paulineribeyre: in fence i do: `check redis; if not in redis: call indexd` so i think we could skip checking redis again here. 3 cases:
- request to indexd from another service: we can assume the other service already checked the cache
- request from a user: now that the request reached indexd, it's too late to improve performance
  - the record is in redis: getting from redis instead of the DB doesn't improve performance (i think)
  - it's not in redis: performance is decreased since indexd checks both

PR #279, opened 2020-06-03
(PXP-1745): Update acl/authz across all versions

mpingram:
Jira Ticket: [PXP-1745](https://ctds-planx.atlassian.net/browse/PXP-1745)

Adds a new endpoint to indexd, PUT `/index/{GUID}/versions`, allowing user to update the `acl` and `authz` fields across all versions of a record.

### New Features
- Added PUT `/index/{GUID}/versions` endpoint, allowing user to update the `acl` and `authz` fields across all versions of a record.

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -504,7 +504,7 @@ def get_all_index_record_versions(record):
     return flask.jsonify(ret), 200
 
 @blueprint.route("/index/<path:record>/versions", methods=["PUT"])
-def get_all_index_record_versions(record):
+def update_all_index_record_versions(record):
     """
     Update metadata for all record versions.
     NOTE currently the only fields that can be updated for all versions are
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -503,6 +503,7 @@ def get_all_index_record_versions(record):
 
     return flask.jsonify(ret), 200
 
+
 @blueprint.route("/index/<path:record>/versions", methods=["PUT"])
 def update_all_index_record_versions(record):
     """
@@ -520,11 +521,7 @@ def update_all_index_record_versions(record):
     acl = request_json.get("acl", None)
     authz = request_json.get("authz", None)
     # authorization and error handling done in driver
-    ret = blueprint.index_driver.update_all_versions(
-        record,
-        acl=acl,
-        authz=authz,
-    )
+    ret = blueprint.index_driver.update_all_versions(record, acl=acl, authz=authz)
 
     return flask.jsonify(ret), 200
 
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -14,6 +14,7 @@ from indexd.errors import UserError
 from .schema import PUT_RECORD_SCHEMA
 from .schema import POST_RECORD_SCHEMA
 from .schema import RECORD_ALIAS_SCHEMA
+from .schema import UPDATE_ALL_VERSIONS_SCHEMA
 
 from .errors import NoRecordFound
 from .errors import MultipleRecordsFound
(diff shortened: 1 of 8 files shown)


indexd/index/blueprint.py
```diff
@@ -503,6 +504,28 @@ def get_all_index_record_versions(record):
...
+    """
+    request_json = flask.request.get_json(force=True)
+    try:
+        jsonschema.validate(request_json, UPDATE_ALL_VERSIONS_SCHEMA)
+    except jsonschema.ValidationError as err:
+        logger.warn(f"Bad request body:\n{err}")
+        raise UserError(err)
+
+    acl = request_json.get("acl", None)
+    authz = request_json.get("authz", None)
```
paulineribeyre, line 522: you can just use `.get("acl")` because None is the default returned value 👍

indexd/index/drivers/alchemy.py
```diff
@@ -1235,6 +1235,59 @@ def get_all_versions(self, did):
...
+                .order_by(IndexRecord.created_date.asc())
+                .all()
+            )
+
+            # User requires update permissions for all versions of the record
+            all_resources = []
+            for rec in records:
+                for resource in rec.authz:
+                    all_resources.append(resource)
+            auth.authorize("update", all_resources)
```
paulineribeyre, line 1270: you can use set comprehension
```
all_resources = {
    resource
    for rec in records
    for resource in rec.authz
}
auth.authorize("update", list(all_resources))
```

tests/test_client.py
```diff
@@ -1776,6 +1777,207 @@ def test_get_all_versions(client, user):
...
+    """
+    If user does not have the 'update' permission on any record, request should
+    fail with 403.
+    """
+    # SETUP
+    # -------
+    admin_user = user
+    test_user = {
+        "Authorization": (
+            "Basic " + base64.b64encode(b"test_user:test").decode("ascii")
```
paulineribeyre, line 1951: indexd can check authorization 2 ways:
- basic auth: if the username and password are correct, then the user is admin, can update anything
- token auth: uses a list of resources to ask arborist if the user has access

here you have 2 users that both use basic auth, so the resources they have access to in arborist should not matter - this test works because you mocked the whole `auth.authorize` function, but the logic is incorrect

you can remove the `use_mock_authz` fixture and just do one call with `user` fixture and one without, to make sure access is checked, or add more complex jwt/arborist mocking :/

indexd/index/blueprint.py
mpingram: Word, thanks! More python knowledge 🐍

tests/test_client.py
mpingram: Gotchu, I think what I'll do if it makes sense is use a JWT/bearer token in the header and keep the mock auth.authorize function -- I wanted to test the specific behavior of 'user has access to one record but not both' which I couldn't figure out how to do with basic auth.

tests/test_client.py
```diff
@@ -1926,53 +1926,44 @@ def test_update_all_versions_fail_on_bad_metadata(client, user):
...
-        {
-            "test": {
-                "create": ["resource_A", "resource_B"],
-                "update": ["resource_A", "resource_B"],
-            },
-            "test_user": {"update": ["resource_B"]},
-        }
-    ],
-)
 def test_update_all_versions_fail_on_missing_permissions(client, user, use_mock_authz):
```
paulineribeyre, line 1929: i think you don't need the `user` fixture at all since everything is mocked 😅

mpingram: Word, I was keeping it in there as a way to insert the first two setup records. I could also do it with a call to use_mock_authz to allow uploading/creating both record_A and record_B.

paulineribeyre: ah got it

PR #282, opened 2020-06-11
(feat) remove presigned_url from drs object

BinamB:

### New Features
- Remove presigned url from the `access_url` field in the drs object when authz is given.

### DIFF
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -50,7 +50,6 @@ def get_signed_url(object_id, access_id):
 
 
 def indexd_to_drs(record, list_drs=False):
-    bearer_token = flask.request.headers.get("AUTHORIZATION")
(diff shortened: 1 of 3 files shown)


indexd/drs/blueprint.py
```diff
@@ -50,7 +50,6 @@ def get_signed_url(object_id, access_id):
 
 
 def indexd_to_drs(record, list_drs=False):
-    bearer_token = flask.request.headers.get("AUTHORIZATION")
```
jiaqi216, line 53: So we still have to accept the auth token but we're just going to ignore it? cc @Avantol13 if you're cool with it then I"m not picky about it either.

BinamB: Oh that's been removed .

Avantol13: yeah, that's fine by me

jiaqi216:
Belated LGTM from me too!

PR #283, opened 2020-06-11
(PXP-6233): fix error message handling for 401 error

BinamB:

### Bug Fixes
- The `object/<guid>/access/<access type>` endpoint throws a 500 instead of a 401 when the user is logged in but does not have access to the specific record.
- Added a general Fence Error handler when drs fails to catch error from fence

### DIFF
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -107,7 +107,7 @@ def handle_user_error(err):
 
 
 @blueprint.errorhandler(AuthzError)
-def handle_auth_error(err):
+def handle_authz_error(err):
     ret = {"msg": str(err), "status_code": 401}
     return flask.jsonify(ret), 401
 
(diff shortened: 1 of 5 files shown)


indexd/drs/blueprint.py
```diff
@@ -106,7 +106,7 @@ def handle_user_error(err):
     return flask.jsonify(ret), 400
 
 
-@blueprint.errorhandler(AuthError)
+@blueprint.errorhandler(AuthzError)
 def handle_auth_error(err):
```
paulineribeyre, line 110:
```suggestion
def handle_authz_error(err):
```

indexd/fence_client.py
```diff
@@ -40,6 +40,8 @@ def get_signed_url_for_object(self, object_id, access_id):
             raise IndexNoRecordFound(
                 "No document with id:{} with access_id:{}".format(object_id, access_id)
             )
+        if req.status_code == 401:
+            raise AuthzError("Unauthorized: Access denied due to invalid credentials.")
         elif req.status_code != 200:
             raise UnexpectedError(req.text)
```
paulineribeyre, line 46: add a log here before raising 500, with status code and details?
unless you're just deleting this sign URL creation soon

PR #290, opened 2020-09-25
feat(mintguid): add new endpoint to mint a valid guid for this instance

Avantol13:
### New Features
- new /mint/guid endpoint to return a valid GUID for this instance without creating a new record


openapis/swagger.yaml
```diff
@@ -1391,6 +1393,38 @@ paths:
            $ref: '#/definitions/URLsOutput'
        400:
          description: invalid request parameters
+  '/guid/mint':
+    get:
+      tags:
+        - GUID
+      summary: search index records by urls metadata key and value
```
paulineribeyre, line 1400: this needs to be updated

indexd/guid/blueprint.py
```diff
@@ -0,0 +1,60 @@
...
+    try:
+        count = int(count)
+    except Exception:
+        return f"Count {count} is not a valid integer", 400
+
+    # clip value on 0, max_count
+    if count < 0:
+        count = 0
+    elif count > max_count:
+        count = max_count
```
paulineribeyre, line 28: maybe we should return an error for these 2 cases? or at least when count > max_count instead, so that users expecting a specific number of GUIDs don't run into unexpected errors?

openapis/swagger.yaml
```diff
@@ -1391,6 +1393,38 @@ paths:
...
+      operationId: getGUIDs
+      produces:
+        - application/json
+      parameters:
+        - in: query
+          name: count
+          type: string
+          description: Number of GUIDs to return, default is 1 (note that there is a max)
```
paulineribeyre, line 1408: should we document what the max is?

tests/default_test_settings.py
```diff
@@ -3,10 +3,6 @@
 from indexd.default_settings import *
 from indexd.index.drivers.alchemy import SQLAlchemyIndexDriver
 
-
-# override the default settings for INDEX because we want to test
-# both PREPEND_PREFIX and ADD_PREFIX_ALIAS, which should not both
-# be set to True in production environments
 CONFIG["INDEX"] = {
```
paulineribeyre, line 6: since we're using the default value now you could just remove this block

tests/default_test_settings.py
```diff
@@ -15,7 +11,7 @@
         index_config={
             "DEFAULT_PREFIX": "testprefix:",
             "PREPEND_PREFIX": True,
-            "ADD_PREFIX_ALIAS": True,
+            "ADD_PREFIX_ALIAS": False,
```
paulineribeyre, line 14: i think [this test](https://github.com/uc-cdis/indexd/blob/f2557f5/tests/test_client.py#L1885-L1902) is not testing the `ADD_PREFIX_ALIAS` feature anymore, it's probably just passing because `PREPEND_PREFIX` is True... but we need to test this feature because it's used in prod

maybe that's not in the scope of this PR

tests/test_guid.py
```diff
@@ -0,0 +1,84 @@
...
+    # result in that many results
+    assert result_count == max(0, count)
+
+
+def test_guids_invalid_count(app, client, user):
+    """
+    Test that generating GUIDs doesn't work when provided invalid count
+    """
+    response = client.get(f"/guid/mint?count=foobar")
+    assert response.status_code != 200
```
paulineribeyre, line 51: i'd test for 400 explicitly because we don't want to return 500s

tests/default_test_settings.py
Avantol13: oh good point, I'll try to fix this

paulineribeyre (approved):

paulineribeyre (approved):

PR #291, opened 2020-10-01
(PXP-6604) DRS endpoint to resolve regardless of prefix

olivarity:
Jira Ticket: [PXP-6604](https://ctds-planx.atlassian.net/browse/PXP-6604)

 Now when attempting to GET a record for a specified GUID, the following resolution steps will be taken:

1. Start by trying to resolve the provided ID
2. If unsuccessful, check what the `DEFAULT_PREFIX` Indexd config is set to
3. If the ID starts with the prefix, attempt to remove the prefix from the provided ID and try resolving again
4. If the ID does not start with the prefix, attempt to add the prefix to the provided ID and try resolving again

### New Features

- Added new `get_with_nonstrict_prefix` helper to `index_driver`.
- Updated `/ga4gh/drs/v1/objects/<path:object_id>` and `/index/<path:object_id>` GET endpoints to utilize this new method.

### DIFF
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -248,7 +248,7 @@ def get_index_record(record):
     Returns a record.
     """
 
-    ret = blueprint.index_driver.get(record)
+    ret = blueprint.index_driver.get_with_nonstrict_prefix(record)
 
     return flask.jsonify(ret), 200
 
--- a/indexd/index/blueprint.py
+++ b/indexd/index/blueprint.py
@@ -717,7 +717,7 @@ def get_bundle_record_with_id(bundle_id):
 
     expand = True if flask.request.args.get("expand") == "true" else False
 
-    ret = blueprint.index_driver.get(bundle_id)
+    ret = blueprint.index_driver.get_with_nonstrict_prefix(bundle_id)
 
     ret = bundle_to_drs(ret, expand=expand, is_content=False)
 
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -1,6 +1,7 @@
 import datetime
 import uuid
 import json
+import re
 from contextlib import contextmanager
 from cdislogging import get_logger
 from sqlalchemy import (
@@ -1108,7 +1109,7 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
             record = self.get(did, expand=expand)
         except NoRecordFound as e:
             DEFAULT_PREFIX = self.config.get("DEFAULT_PREFIX")
-            match = re.match(r"(dg\.[0-9a-f]{4}\/)(.+)", object_id)
+            match = re.match(r"(dg\.[0-9a-f]{4}\/)(.+)", did)
             if match and match.group(1) == DEFAULT_PREFIX:
                 record = self.get(match.group(2), expand=expand)
             elif match is None:
@@ -1116,7 +1117,7 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
             else:
                 raise e
 
-        return ret
+        return record
 
     def update(self, did, rev, changing_fields):
         """
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -1107,14 +1107,17 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
             record = self.get(did, expand=expand)
         except NoRecordFound as e:
             DEFAULT_PREFIX = self.config.get("DEFAULT_PREFIX")
-            if "/" in did:
-                prefix, uuid = did.rsplit("/", 1)
-                if prefix + "/" == DEFAULT_PREFIX:
-                    record = self.get(uuid, expand=expand)
+            if DEFAULT_PREFIX is not None:
+                if "/" in did:
+                    prefix, uuid = did.rsplit("/", 1)
+                    if prefix + "/" == DEFAULT_PREFIX:
+                        record = self.get(uuid, expand=expand)
+                    else:
+                        raise e
                 else:
-                    raise e
+                    record = self.get(DEFAULT_PREFIX + did, expand=expand)
             else:
-                record = self.get(DEFAULT_PREFIX + did, expand=expand)
+                raise e
 
         return record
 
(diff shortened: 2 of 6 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -1096,6 +1097,28 @@ def get(self, did, expand=True):
...
+        Proxies 'get' with provided id.
+        If not found but prefix matches default, attempt with prefix stripped.
+        If not found and id has no prefix, attempt with default prefix prepended.
+        """
+        try:
+            # Try to resolve id as provided
+            record = self.get(did, expand=expand)
+        except NoRecordFound as e:
+            DEFAULT_PREFIX = self.config.get("DEFAULT_PREFIX")
+            match = re.match(r"(dg\.[0-9a-f]{4}\/)(.+)", did)
```
olivarity, line 1112: @paulineribeyre do we only want to attempt resolution if the prefix matches this specific format? or should we just see if the provided id starts with whatever `DEFAULT_PREFIX` is set to, even if it's not spec compliant?

paulineribeyre:
could you move the non-bullet points part of the PR description to before "New Features"? 🙏  our script that parses PR descriptions to generate release notes interprets each new line as a "new feature"

indexd/index/blueprint.py
```diff
@@ -248,7 +248,7 @@ def get_index_record(record):
     Returns a record.
     """
 
-    ret = blueprint.index_driver.get(record)
+    ret = blueprint.index_driver.get_with_nonstrict_prefix(record)
```
paulineribeyre, line 251: i think we should also call `get_with_nonstrict_prefix` instead of `get` in `get_bundle_record_with_id` and `indexd/blueprint.get_record`. maybe even in `get_dos_record` for consistency, although that one is not used anymore

indexd/index/drivers/alchemy.py
paulineribeyre: i would limit it to the configured `DEFAULT_PREFIX` (we call it default, but i'm pretty sure it's actually the only allowed prefix)

indexd/index/drivers/alchemy.py
```diff
@@ -1096,6 +1097,28 @@ def get(self, did, expand=True):
...
+            if match and match.group(1) == DEFAULT_PREFIX:
+                record = self.get(match.group(2), expand=expand)
+            elif match is None:
```
paulineribeyre, line 1115: i think something like this could let us avoid the regex
```
if "/" in did: (because we use UUIDs, there's only a `/` if there's a prefix)
    if did.split("/")[0] == DEFAULT_PREFIX:
        try without prefix
else:
    try with prefix
```
and then of course we can have a test case with a GUID that exists, but a prefix different from DEFAULT_PREFIX, and make sure we return 404

indexd/index/blueprint.py
olivarity: Sorry, just realizing I missed this after the first review. I can make these changes, but I'm not familiar with what those methods do; would this require any additional testing, or should the existing cases cover it?

paulineribeyre: - `indexd.blueprint.get_record`: instead of only testing the `https://.../index/<id>` endpoint, the `https://.../index/index/<id>` should also be tested
- `get_bundle_record_with_id`: yeah testing this one would require creating a bundle first
- `get_dos_record` let's not update it, better not break legacy code

olivarity: @paulineribeyre regarding the first point, wouldn't it be the other way around? `https://.../index/index/<id>` is the one being updated to the use the new method in this PR; `https://.../index/<id>` hasn't been touched, and from the background info I thought we wanted to leave that one as-is.

olivarity: I think I see the confusion, my comments in the PR body are from the perspective of `/` being the indexd root, not the entire deployment.

paulineribeyre: right sorry, the other way around

why from the background info?

paulineribeyre: missing `indexd.blueprint.get_record`

tests/test_driver_alchemy_crud.py
```diff
@@ -494,6 +494,161 @@ def test_driver_get_fails_with_no_records():
...
+        """,
+            ("testprefix/" + did, baseid, rev, form, size, created_date, updated_date),
+        )
+
+        conn.commit()
+
+        record = driver.get_with_nonstrict_prefix(did)
+
+        assert record["did"] == "testprefix/" + did, "record id does not match"
+        assert record["baseid"] == baseid, "record id does not match"
```
paulineribeyre, line 534:
```suggestion
        assert record["baseid"] == baseid, "record baseid does not match"
```

tests/test_driver_alchemy_crud.py
```diff
@@ -494,6 +494,161 @@ def test_driver_get_fails_with_no_records():
...
+        """,
+            (did, baseid, rev, form, size, created_date, updated_date),
+        )
+
+        conn.commit()
+
+        record = driver.get_with_nonstrict_prefix("testprefix/" + did)
+
+        assert record["did"] == did, "record id does not match"
+        assert record["baseid"] == baseid, "record id does not match"
```
paulineribeyre, line 584:
```suggestion
        assert record["baseid"] == baseid, "record baseid does not match"
```

indexd/index/drivers/alchemy.py
```diff
@@ -1096,6 +1096,31 @@ def get(self, did, expand=True):
 
             return record.to_document_dict()
 
+    def get_with_nonstrict_prefix(self, did, expand=True):
```
paulineribeyre, line 1099: could you add 1 or 2 sentences at the end of [this section](https://github.com/uc-cdis/indexd#distributed-resolution-utilizing-prefixes-in-guids) to mention which endpoints resolve regardless of prefix usage, and that for other endpoints you need to get the real `did`

indexd/index/blueprint.py
olivarity: done, testing

indexd/index/drivers/alchemy.py
olivarity: added

paulineribeyre (approved):

PR #293, opened 2020-10-16
add _dist endpoint

mattreex:
Jira Ticket: [PXP-7004](https://ctds-planx.atlassian.net/browse/PXP-7004)

### New Features
Implemented dist endpoint and added unit test

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -2575,6 +2575,4 @@ def test_version_check(client):
 
 def test_get_dist(client):
     res = client.get("/_dist")
-    if not res.json:
-        assert res.status_code == 404
     assert res.status_code == 200
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -2572,3 +2572,8 @@ def test_status_check(client):
 def test_version_check(client):
     res = client.get("/_version")
     assert res.status_code == 200
+
+
+def test_get_dist(client):
+    res = client.get("/_dist")
+    assert res.status_code == 200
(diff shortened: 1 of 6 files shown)


tests/test_client.py
```diff
@@ -2572,3 +2573,10 @@ def test_status_check(client):
 def test_version_check(client):
     res = client.get("/_version")
     assert res.status_code == 200
+
+
+def test_get_dist(client):
+    res = client.get("/_dist")
+    if not res.json:
+        assert res.status_code == 404
+    assert res.status_code == 200
```
paulineribeyre, line 2582: unit tests should always have the same behavior

ideally we'd want 2 separate test cases:
1. there is a dist config, return it => you can add a dist config to the [test settings](https://github.com/uc-cdis/indexd/blob/master/tests/default_test_settings.py)
2. there is no dist config, return empty list => to test this, you can have a look at how to use a different config than the default. if we can't, only checking case 1 should be okay

tests/test_client.py
```diff
@@ -2572,3 +2572,8 @@ def test_status_check(client):
 def test_version_check(client):
     res = client.get("/_version")
     assert res.status_code == 200
+
+
+def test_get_dist(client):
+    res = client.get("/_dist")
+    assert res.status_code == 200
```
paulineribeyre, line 2579: let's `assert` the contents of the response are as expected, too (not that helpful for the current code, but will be if we reformat)

paulineribeyre (approved):

PR #294, opened 2020-10-22
fix(updated): actually update updated_date when modifying record

Avantol13:
### Bug Fixes
- update updated_date when modifying record

### DIFF
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -878,6 +878,8 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
 
             record.rev = str(uuid.uuid4())[:8]
 
+            record.updated_date = datetime.datetime.utcnow()
+
             session.add(record)
             session.commit()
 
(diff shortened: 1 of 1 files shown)


indexd/index/drivers/alchemy.py
```diff
@@ -878,6 +878,8 @@ def update_blank_record(self, did, rev, size, hashes, urls, authz=None):
 
             record.rev = str(uuid.uuid4())[:8]
 
+            record.updated_date = datetime.datetime.utcnow()
```
paulineribeyre, line 881: all other dates in indexd use `.now()` so if you want to use `.utcnow()` maybe we should change it everywhere? 😕

paulineribeyre: oh nvm, i'm wrong, only unit tests use `.now()`

paulineribeyre (approved):

indexd/index/drivers/alchemy.py
Avantol13: that's what it uses when it creates the record in the db :grimacing: so I went based off that

Avantol13: at least, it's the default https://github.com/uc-cdis/indexd/blob/500238313f7439c9a15f78533b947c89470fad1a/indexd/index/drivers/alchemy.py#L75

PR #297, opened 2020-11-10
Fix authz check for empty authz

paulineribeyre:
### Improvements
- Records with an empty `authz` can now be updated by admins

### DIFF
--- a/indexd/auth/drivers/alchemy.py
+++ b/indexd/auth/drivers/alchemy.py
@@ -131,8 +131,8 @@ class SQLAlchemyAuthDriver(AuthDriverABC):
                 "Arborist is not configured; cannot perform authorization check"
             )
         if not resource:
-            # TODO: fix this. Setting authz to [] throws this error but
-            # admins should be able to do it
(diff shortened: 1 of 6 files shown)


indexd/auth/drivers/alchemy.py
```diff
@@ -131,8 +131,8 @@ def authz(self, method, resource):
                 "Arborist is not configured; cannot perform authorization check"
             )
         if not resource:
-            # TODO: fix this. Setting authz to [] throws this error but
-            # admins should be able to do it
-            raise AuthError("Permission denied.")
+            # if the `authz` is empty, admins should still be able to perform
+            # operations on the record. For now, admin = access to `/programs`
+            resource = ["/programs"]
```
paulineribeyre, line 136: `"*"` => arborist returns false - interprets it as a resource
`"/"` => arborist returns a "missing resource in auth request" error

vpsx (approved):
for reference https://cdis.slack.com/archives/CEDUCCCM7/p1605044691171500

PR #302, opened 2021-01-12
Fix/expand

giangbui:
Jira Ticket: [PXP-7386](https://ctds-planx.atlassian.net/browse/PXP-7386)

### New Features
- Show first layer of bundles when expand=false


BinamB (approved):
lgtm!

PR #303, opened 2021-02-08
PXP-7557: Propagate params in DRS signed url endpoint 

mcannalte:
### Bug Fixes

- all url query parameters sent to the DRS signed url endpoint now propagate to the corresponding Fence request (instead of stripping off all query params)

### DIFF
--- a/tests/test_drs.py
+++ b/tests/test_drs.py
@@ -241,9 +241,15 @@ def test_get_presigned_url_with_query_params(client, user):
     rec_1 = res_1.json
     access_id_list = ["s3", "gs", "ftp"]
     for access_id in access_id_list:
-        presigned = generate_presigned_url_response(rec_1["did"], access_id, "k=v")
+        presigned = generate_presigned_url_response(
+            rec_1["did"], access_id, "userProject=someproject&optional=val"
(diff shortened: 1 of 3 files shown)


tests/test_drs.py
```diff
@@ -6,10 +6,23 @@
 from tests.test_bundles import get_bundle_doc
 
 
-def generate_presigned_url_response(did, protocol="", status=200):
-    full_url = (
-        "https://fictitious-commons.io/data/download/" + did + "?protocol=" + protocol
-    )
+def generate_presigned_url_response(did, protocol="", query_string="", status=200):
```
Avantol13, line 9: minor nitpick: protocol ends up in the query string already, so `query_string` is really like `additional_query_param` and it might be better for this function to accept them as a dictionary and then dump them into the right format so the caller doesn't have to query whether or not to include ? or &. basically, standardize this function as `func(did, params, status)`

Avantol13 (changes requested):

Avantol13:
looks good! I left a minor nitpick about some existing structure for the test function interface

Avantol13 (approved):

PR #304, opened 2021-04-06
PXP-6824: return 401 errors from Arborist

MaribelleHGomez:

### Improvements
Uses new image that has proper install of gen3authz package installation of Cryptography requires Rust (https://github.com/uc-cdis/cloud-automation/pull/1568

### Dependency updates

- pybase3-1.5.0

- gen3authz"^1.0.4"

- sqlalchemy pinned to "~1.3.3", SQLAlchemy_utils not supported for 1.4

### Deployment changes

When running gen3authz,  backoff package needed separate poetry run pip install


indexd/auth/drivers/alchemy.py
```diff
@@ -131,5 +131,11 @@ def authz(self, method, resource):
...
             # TODO: Figure out how to handle Gen3 operational admins in a better way
             resource = ["/programs"]
-        if not self.arborist.auth_request(get_jwt_token(), "indexd", method, resource):
-            raise AuthError("Permission denied.")
+
+        try:
+            if not self.arborist.auth_request(
+                get_jwt_token(), "indexd", method, resource
+            ):
+                raise AuthError("Permission denied.")
```
BinamB, line 139: So you dont have to do the if. Look up how try except works. So under the try block if you run a function and it fails then it goes to the except block with whatever error that function spits out. Here's an example of try-except that I've written in fence https://github.com/uc-cdis/fence/blob/master/fence/resources/openid/ras_oauth2.py#L129

indexd/auth/drivers/alchemy.py
```diff
@@ -131,5 +131,11 @@ def authz(self, method, resource):
             # operations on the record. For now, admin = access to `/programs`.
             # TODO: Figure out how to handle Gen3 operational admins in a better way
             resource = ["/programs"]
-        if not self.arborist.auth_request(get_jwt_token(), "indexd", method, resource):
-            raise AuthError("Permission denied.")
+
+        try:
+            if not self.arborist.auth_request(
```
BinamB, line 136: In terms of unit testing I would try to mock `self.arborist.auth_request` to return the 401 error. Pytest is really cool like that. Add a test that calls this `authz` function or whatever the endpoint calls the `authz` function and mock the return for `auth_request`

pyproject.toml
```diff
@@ -16,13 +16,13 @@ cdislogging = "^1.0.0"
...
 dosclient = {git = "https://github.com/uc-cdis/dosclient", rev = "1.1.0"}
-gen3authz = "^0.4.0"
+gen3authz = "^1.0.4"
 hsclient = {git = "https://github.com/uc-cdis/hsclient", rev = "1.0.0"}
 indexclient = "^2.1.0"
 jsonschema = "^2.5.1"
 flask = "^1.1.1"
 psycopg2 = "^2.7"
-sqlalchemy = "^1.3.3"
+sqlalchemy = "~1.3.3"
```
BinamB, line 25: Could you remind me why we're changing this?

MaribelleHGomez:
It's being changed because of compatibility with sqlalchemy_utils! 1.4 does
not support sqlalchemy_utils yet

indexd/auth/drivers/alchemy.py
MaribelleHGomez: Are you sure that arborist will always return an error if someone doesn't have permission, what about where `elif response.code == 200:
            return bool(response.json["auth"])` returns false?

pyproject.toml
MaribelleHGomez: It's being changed because of compatibility with sqlalchemy_utils! 1.4 does
not support sqlalchemy_utils yet

BinamB:
being picky but can you format the PR description properly. Like add bullet points and such. This might be helpful https://guides.github.com/features/mastering-markdown/

indexd/index/blueprint.py
```diff
@@ -780,6 +780,11 @@ def handle_auth_error(err):
     return flask.jsonify(error=str(err)), 403
 
 
+@blueprint.errorhandler(AuthzError)
+def handle_authz_error(err):
+    return flask.jsonify(error=str(err)), 401
+
```
BinamB, line 786: What it was thinking was instead of it being an AuthZ error you can make it some general arborist error where you can pass the error message + error code to this function and do something like `return flask.jsonify(error=err_msg), err_code`

BinamB (approved):
lgtm i guess!

indexd/index/blueprint.py
MaribelleHGomez: I think the decision to make it an auth error was because regardless of it was a failure of the arborist service or a token expiration, the authorization process still failed

PR #305, opened 2021-05-03
PXP-7499: Resolve regardless of the prefix when using /index?ids

MaribelleHGomez:
### Improvements

- Modified get_with_nonstrict prefix route (https://github.com/uc-cdis/indexd/blob/master/indexd/index/drivers/alchemy.py#L1101) to replace hardcoded prefix delimiter.
- (https://github.com/uc-cdis/indexd/blob/master/indexd/index/drivers/alchemy.py#L500) expanded logic to query additional ids by adding to id array. If the default prefix is not _None_ the id with default prefix will be appended to id array; similarly, following logic from the get_with_nonstrict prefix route, an id stripped of the default prefix will be added to the id array.
- Added test for get without prefix (https://github.com/uc-cdis/indexd/blob/chore/update_id_get/tests/test_client.py#L86)


MaribelleHGomez:
I ran tests locally but I'd like some more feedback on unit tests

BinamB:
> I ran some tests locally but I'd like some more feedback on unit tests
> These are the two tests for that area of code:
> `data_by_ids = client.get("/index/?ids={}".format(rec_1["did"])) assert data_by_ids.status_code == 200 data_list_all = data_by_ids.json`
> 
> ```
> `# with nonstrict prefix
> stripped = rec_1["did"].split("testprefix:", 1)[1]
> data_by_ids = client.get("/index/?ids={}".format(stripped))
> assert data_by_ids.status_code == 200
> data_list_all = data_by_ids.json`
> ```
> 
> also, some feedback on this (https://github.com/uc-cdis/indexd/blob/chore/update_id_get/indexd/index/drivers/alchemy.py#L511), ty

tests/test_client.py
```diff
@@ -83,6 +83,12 @@ def test_index_list_with_params(client, user):
     assert data_by_ids.status_code == 200
     data_list_all = data_by_ids.json
```
BinamB, line 84: This isn't being tested anymore since in ln 90 `data_list_all` replaces this one.

BinamB: So i  guess i would suggest add the same types of checks in line 92 over here.

tests/test_client.py
```diff
@@ -83,6 +83,12 @@ def test_index_list_with_params(client, user):
...
 
+    # with nonstrict prefix
+    stripped = rec_1["did"].split("testprefix:", 1)[1]
+    data_by_ids = client.get("/index/?ids={}".format(stripped))
+    assert data_by_ids.status_code == 200
+    data_list_all = data_by_ids.json
+
```
BinamB, line 91: I guess I would also add different types of test cases like the one we talked about in chat where a guid and a pref+guid might exist. That is something we definitely want to handle.

BinamB (changes requested):
I think its good. Just a small comment. Could you also remind me what Pauline wanted changed as well?

BinamB (approved):

BinamB (approved):

indexd/index/drivers/alchemy.py
```diff
@@ -498,7 +498,25 @@ def ids(
                 query = query.order_by(IndexRecord.did)
 
             if ids:
-                query = query.filter(IndexRecord.did.in_(ids))
+                DEFAULT_PREFIX = self.config.get("DEFAULT_PREFIX")
+                found_ids = []
+                new_ids = []
+
+                if not DEFAULT_PREFIX:
+                    print("NO DEFAULT_PREFIX")
```
BinamB, line 506: can you log this instead of print

BinamB (changes requested):

BinamB (approved):

PR #309, opened 2021-06-04
Bundles: don't swallow exception + Old indexd_admin deprecation log

paulineribeyre:
### Improvements
- Bundles endpoints: don't swallow authorization exceptions
- Add a deprecation log to the old "indexd_admin" logic


MaribelleHGomez (approved):
I like the changes you made in blueprint so it doesn't return same err for everything

PR #311, opened 2021-06-07
PXP-6245: update packages for nested dependency version resolution

MaribelleHGomez:
Jira Ticket: [PXP-6245](https://ctds-planx.atlassian.net/browse/PXP-6245)

An update for jsonschema was needed to utilize sheepdog for poetry, updated other packages as well to maintain relevancy.

### Dependency updates
Update versions: python, authutils, cdiserrors, jsonschema, flask, pandas, sqlalchemy-utils, coveralls, pytest, pytest-cov, pytest-flask, and responses

http://json-schema.org/understanding-json-schema/reference/schema.html ref to change regarding $schema uri


pyproject.toml
```diff
@@ -10,31 +10,32 @@ include = [
...
-python = "3.6.*"
-authutils = "^4.0.0"
+python = "^3.6.1"
+authutils = "^6.0.0"
 cdislogging = "^1.0.0"
-cdiserrors =  "^0.1.2"
+cdiserrors =  "^1.0.0"
 doiclient = {git = "https://github.com/uc-cdis/doiclient", rev = "1.0.0"}
 dosclient = {git = "https://github.com/uc-cdis/dosclient", rev = "1.1.0"}
 gen3authz = "^1.0.4"
```
paulineribeyre, line 19: remove file `pkg/gen3authz-1.0.6.tar.gz`

pyproject.toml
```diff
@@ -10,31 +10,32 @@ include = [
...
 hsclient = {git = "https://github.com/uc-cdis/hsclient", rev = "1.0.0"}
 indexclient = "^2.1.0"
-jsonschema = "^2.5.1"
-flask = "^1.1.1"
+jsonschema = "^3.2"
+flask = "^2.0.1"
+pandas= "^1.0"
```
paulineribeyre, line 24: why do we need pandas?

pyproject.toml
```diff
@@ -10,31 +10,32 @@ include = [
...
+cdiserrors =  "^1.0.0"
 doiclient = {git = "https://github.com/uc-cdis/doiclient", rev = "1.0.0"}
 dosclient = {git = "https://github.com/uc-cdis/dosclient", rev = "1.1.0"}
 gen3authz = "^1.0.4"
 hsclient = {git = "https://github.com/uc-cdis/hsclient", rev = "1.0.0"}
 indexclient = "^2.1.0"
-jsonschema = "^2.5.1"
-flask = "^1.1.1"
+jsonschema = "^3.2"
+flask = "^2.0.1"
```
paulineribeyre, line 23: woah! ✨ 
did you check if the breaking changes from 2.0.0 would affect indexd?

indexd/index/schema.py
```diff
@@ -54,7 +54,7 @@
                 "sha256": {"type": "string", "pattern": "^[0-9a-f]{64}$"},
                 "sha512": {"type": "string", "pattern": "^[0-9a-f]{128}$"},
                 "crc": {"type": "string", "pattern": "^[0-9a-f]{8}$"},
-                "etag": {"type": "string", "pattern": "^[0-9a-f]{32}(-\d+)?$"},
+                "etag": {"type": "string", "pattern": "^[0-9a-f]{32}(-\\d+)?$"},
```
paulineribeyre, line 57: i don't think that's correct, we want `\d` (a digit), not `\\d` (literally the characters "/d"), why did we need this change? 🤔

indexd/index/schema.py
```diff
@@ -167,7 +167,7 @@
                         "if": {"properties": {"type": {"const": "etag"}}},
                         "then": {
                             "properties": {
-                                "checksum": {"pattern": "^[0-9a-f]{32}(-\d+)?$"}
+                                "checksum": {"pattern": "^[0-9a-f]{32}(-\\d+)?$"}
```
paulineribeyre, line 170: same

indexd/alias/schema.py
```diff
@@ -32,7 +32,7 @@
                 "sha256": {"type": "string", "pattern": "^[0-9a-f]{64}$"},
                 "sha512": {"type": "string", "pattern": "^[0-9a-f]{128}$"},
                 "crc": {"type": "string", "pattern": "^[0-9a-f]{8}$"},
-                "etag": {"type": "string", "pattern": "^[0-9a-f]{32}(-\d+)?$"},
+                "etag": {"type": "string", "pattern": "^[0-9a-f]{32}(-\\d+)?$"},
```
paulineribeyre, line 35: same

pyproject.toml
MaribelleHGomez: Well it passed jenkins and travis, is there anything else I should double check?
Looking at https://flask.palletsprojects.com/en/2.0.x/changes/#version-2-0-0 some of the major changes like dep support for python <=3.5 and bump up some other minimum req should be fine with what we've got.
There re some json func dep but we don't use them

pyproject.toml
MaribelleHGomez: ha I can't find this file?

pyproject.toml
MaribelleHGomez: because of cython, which we needed for... I can't remember but I can try and recreate if you want

pyproject.toml
paulineribeyre: it's in your PR files https://github.com/uc-cdis/indexd/pull/311/files 🤷‍♀️

pyproject.toml
paulineribeyre: that's fine but could you add a comment on why we need it 😬 i know it's annoying to wait for jenkins again but context is nice when we update these files - maybe use a label to limit jenkins to a small set of tests to speed it up

pyproject.toml
paulineribeyre: I wish I could say I trust jenkins 100% but 😆 you never know
But yeah if you read through the list and you think it's fine, it's all good

pyproject.toml
MaribelleHGomez: it says it's removed though right?when i hover over the red part it says binary file removed

paulineribeyre: ugh sorry, let's just say i'm tired

MaribelleHGomez: no you're good, I thought it was definitely me because I had it in another branch too ._.

pyproject.toml
MaribelleHGomez: in the toml right? 😳

paulineribeyre: yes

indexd/index/schema.py
MaribelleHGomez: I think you're right, I changed it because of a warning about future deprecation, but now it's back to how it was before, we'll just need to change it in a future upgrade for python

https://docs.python.domainunion.de/3/library/re.html

paulineribeyre (approved):

PR #315, opened 2021-07-13
Add local development notes

plooploops:
### Improvements
* Add notes for local development

### DIFF
--- a/docs/local_dev_environment.md
+++ b/docs/local_dev_environment.md
@@ -1,0 +1,238 @@
+# Set up a local development environment
+
+This guide will cover setting up an indexd development environment.
+
+## Set up Working Directory
+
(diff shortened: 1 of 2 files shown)


docs/local_dev_environment.md
```diff
@@ -0,0 +1,219 @@
+# Set up a local development environment
```
williamhaley, line 1: I believe the current `indexd` `README.md` has some instructions already. I think you've made it so that this file covers almost everything in the `README.md`, but can you delete the now-redundant `README.md` info and link to this file from there?

plooploops: Updated and thanks for the feedback.  Let me know if this works.

williamhaley (approved):

PR #325, opened 2022-01-21
feat(python): python 3.10 upgrade

SpencerAxelrod:
Jira Ticket: [PXP-9243](https://ctds-planx.atlassian.net/browse/PXP-9243)

### Improvements
- Upgrade to Python 3.10


Avantol13 (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+        ports:
+          # Maps tcp port 5432 on service container to the host
+          - 5432:5432
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.10.0
+    - uses: actions/cache@preview
```
SpencerAxelrod, line 30: Not entirely sure how/where this cache action is used or if it is needed

Avantol13: this is caching the virtual env so if the same files come in again without any dependency changes, it doesn't have to reinstall everything - so we should keep it

pyproject.toml
```diff
@@ -10,7 +10,7 @@ include = [
 ]
```
Avantol13, line 10: let's update the `version` above to a minor semver update. So 3.2.0 and then we can do a service release after this

.secrets.baseline
```diff
@@ -62,7 +62,7 @@
       {
         "hashed_secret": "8c0fdc72bf4daf2a2338287334d0cbd221fa9ef4",
         "is_verified": false,
-        "line_number": 28,
+        "line_number": 30,
```
paulineribeyre, line 65: the secrets baseline is outdated, since you deleted `.travis.yml` it shouldn't be in here anymore

SpencerAxelrod: thanks pauline!

PR #335, opened 2022-09-06
Update error message

paulineribeyre:
indexd's GET endpoint checks if a record exists and if not, if a bundle exists. If neither exist, we currently return the error message "no bundle found" that comes from `get_bundle()`. We should return "no record found" instead since this endpoint is not specific to bundles

### Improvements
- Update `GET /index/<GUID>` error message


MaribelleHGomez (approved):
:) lgtm

PR #336, opened 2022-09-19
fix(migrate): improve acl->authz migration, remove deprecated endpoints

Avantol13:
…fence configuration

### Breaking Changes
- Ensure you are using Fence>=2021.10. Remove deprecated GA4GH access endpoints. these have existed (and been used) from the Fence microservice since last year. See cloud automation change [here](https://github.com/uc-cdis/cloud-automation/blob/f10f04f7957eb2ba179504f7df9d6429f7298835/gen3/bin/kube-setup-revproxy.sh#L78-L79)

### Improvements
- allow passing of fence config for authz mapping to the acl migration script


johnfrancismccann (approved):

Avantol13:
Merged knowing that GetCoreMetadata test is failing. It's unrelated to these changes and a known issue that is being debugged. Likely a portal fix is necessary.

PR #338, opened 2022-09-28
Revert "Revert "fix(migrate): improve acl->authz migration, remove de…

Avantol13:
…precated endpoints (#336)" (#337)"

COPY of #336 with addition of fixing broken integration test. Had previously force merged, reverted, now fixing

### Breaking Changes
- Ensure you are using Fence>=2021.10. Remove deprecated GA4GH access endpoints. these have existed (and been used) from the Fence microservice since last year. See cloud automation change [here](https://github.com/uc-cdis/cloud-automation/blob/f10f04f7957eb2ba179504f7df9d6429f7298835/gen3/bin/kube-setup-revproxy.sh#L78-L79)

### Improvements
- allow passing of fence config for authz mapping to the acl migration script


bin/migrate_acl_authz.py
```diff
@@ -211,11 +242,7 @@ def acl_to_authz(self, record):
                     self.projects[acl_item], acl_item
                 )
                 projects_found += 1
-            else:
-                # nothing worked, raise exception
-                raise EnvironmentError(
-                    "program or project {} does not exist".format(acl_item)
```
paulineribeyre, line 217: should we keep this at least as a warning log?

paulineribeyre:
i think we should add the changes to the PR description even if they have already been published in release notes

bin/migrate_acl_authz.py
Avantol13: the warning would end up needing to be more generic b/c now it's not just that program/proj don't exist, but that there's no mapping either. I think the error log in a few lines sorta covers that? https://github.com/uc-cdis/indexd/pull/338/files/24af7553837da98ddf5a53259b91e4f87d23a936#diff-02f08e0cd101b79f6a25f6d6689008de374d7b6399a2a1291ee52faafc73aacdR247

Lemme know if you disagree

Avantol13: basically if nothing is found in any of the cases then there'll be an error log

paulineribeyre: ah ok, yeah that other log works

paulineribeyre (approved):

Avantol13:
Force merging even though a single test is failing. It is a known faulty test

PR #346, opened 2023-02-21
(PPS-196): Fix Snyk Complaining about Code Injection

BinamB:

### Bug Fixes
- Change using `eval` to `json.loads` to prevent code injections.

### Improvements
- Sanitize user input for additional configs.

### DIFF
--- a/bin/config_helper.py
+++ b/bin/config_helper.py
@@ -176,6 +176,14 @@ def _get_all_additional_configs(other_files):
     """
     additional_configs = dict()
     for file_path in other_files:
+        # Sanitize user input to ensure file_path is a valid Python or json file
+        if not os.path.isfile(file_path) or (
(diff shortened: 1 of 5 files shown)


bin/config_helper.py
```diff
@@ -176,6 +176,14 @@ def _get_all_additional_configs(other_files):
     """
     additional_configs = dict()
     for file_path in other_files:
+        # Sanitize user input to ensure file_path is a valid Python or json file
```
paulineribeyre, line 179: The only `config_helper` function we use in indexd is `load_json`. So I think you can remove these fence-specific functions. That's what i did in Sheepdog and Peregrine, for example https://github.com/uc-cdis/sheepdog/blob/923f796/bin/confighelper.py

paulineribeyre (approved):

PR #347, opened 2023-02-23
PXP-10714: Prefer version to rev

MaribelleHGomez:
Currently the DRS response version prefers indexd rev instead of version.
This prefers version before rev. If version is not available it should fallback to rev.

### Bug Fixes
- DRS response prefers version before rev, if version is not available it uses rev


BinamB:
Could you add what you fixed as a point under bug fixes in the PR description?

BinamB (approved):

PR #355, opened 2023-03-30
Feat/pull bucket info

BinamB:

### New Features
- Add bucket region information to DRS object


Avantol13 (changes requested):
This caching strategy isn't super useful when there are many instances of the service running (since all the caches are in-memory and specific to that pod). Since we expect many instances of indexd to be running, I would suggest we look into a more centralized caching solution similar to what's been done in Fence (look for how we dealt with caching creds during signed URL calls or even some of RAS stuff maybe). The idea being: we need a central place for newly spun up instances of indexd to get the cached information. Naive solution (and what we did in Fence) was have a new database table to store the central cache AND in-mem cache in the instances

tests/test_aliases_endpoints.py
```diff
@@ -1,3 +1,5 @@
+import os
```
k-burt-uch, line 1: Unused import

indexd/app.py
```diff
@@ -3,6 +3,7 @@
 
 from alembic.config import main as alembic_main
 import cdislogging
+from distutils.command.config import config
```
k-burt-uch, line 6: Do you still need this config here?

k-burt-uch (changes requested):

PR #362, opened 2023-07-17
adding gunicorn instead of uwsgi and modifying the base image. 

EliseCastle23:
### Improvements
- Update to use new Amazon Linux base image and use the same structure as our other python services. 
- Utilizing "gen3" user instead of "root" for more secure containers
- Moving to Poetry to manage our virtual environments 
- Multi-stage Docker builds for smaller images
- Move to Gunicorn


Dockerfile
```diff
@@ -1,49 +1,80 @@
-# To run: docker run -v /path/to/wsgi.py:/var/www/indexd/wsgi.py --name=indexd -p 81:80 indexd
-# To check running container: docker exec -it indexd /bin/bash
+ARG AZLINUX_BASE_VERSION=master
 
-FROM quay.io/cdis/python:python3.9-buster-2.0.0
+# Base stage with python-build-base
+FROM quay.io/cdis/python-build-base:${AZLINUX_BASE_VERSION} AS base
+
+# Comment this in, and comment out the line above, if quay is down
+# FROM 707767160287.dkr.ecr.us-east-1.amazonaws.com/gen3/python-build-base:${AZLINUX_BASE_VERSION} as base
```
Avantol13, line 7: is there a reason we're not preferring ECR?

Dockerfile
```diff
@@ -1,49 +1,80 @@
...
+    chown -R gen3:gen3 /venv
+
+
+# Builder stage
+FROM base AS builder
+
+USER gen3
+
+
+RUN python -m venv /venv
```
Avantol13, line 30: so `poetry install` will use a venv automatically when you install. And in fact, we're explicitly setting https://python-poetry.org/docs/configuration/#virtualenvsin-project, so it should be deterministic where the venv gets created. So as long as you run all future commands you want within the env with `poetry run ....` then it should just work.

I don't think we need (or want) any other venv setup or activation outside of the stock poetry support. It could create some confusing differences

Dockerfile
jawadqur: So that community may be able to build/ develop and I belive ECR has a bandwidth cost, while quay is free.

Dockerfile
```diff
@@ -1,49 +1,42 @@
-# To run: docker run -v /path/to/wsgi.py:/var/www/indexd/wsgi.py --name=indexd -p 81:80 indexd
-# To check running container: docker exec -it indexd /bin/bash
+ARG AZLINUX_BASE_VERSION=master
 
-FROM quay.io/cdis/python:python3.9-buster-2.0.0
+## For local development
+# FROM quay.io/cdis/python-nginx-al2:feat_python-nginx AS base
+
+FROM 707767160287.dkr.ecr.us-east-1.amazonaws.com/gen3/python-nginx-al2:feat_python-nginx as base
```
Avantol13, line 6: let's follow the warning and fix `The 'as' keyword should match the case of the 'from' keyword`

docker-compose.yaml
```diff
@@ -0,0 +1,36 @@
+version: '3.8'
```
Avantol13, line 1: per previous conversations, I'd like to remove these separate ideas around local dev improvements to speed up the approval and review of the base image upgrades. We can have a separate PR and discussions to talk through ideas like this

pyproject.toml
```diff
@@ -21,13 +21,13 @@ gen3authz = "^1.0.4"
 hsclient = {git = "https://github.com/uc-cdis/hsclient", rev = "1.0.0"}
 indexclient = "^2.1.0"
 jsonschema = "^3.2"
+psycopg2-binary = "^2.7"
+gunicorn = "^22.0.0"
```
Avantol13, line 25: If we are not blocked from upgrading beyond these versions, I'd like to get in the habit of specifying >= rather than the default ^. If things break, we'll see it in unit tests and either update so the newer lib works, or at that point we'll restrict the versioning and create a ticket to upgrade. 

As it is, we'll never automatically try to update to new major versions.

So, even though the other deps aren't like this _yet_, we could replace all ^ with >= at least with the new deps. 

If you're feeling adventurous, we could try everything else too. With the exception of sqlalchemy, which I know won't work b/c 2.0 requires updates

Dockerfile
Avantol13: ahh, that makes sense 👍

Avantol13: is that acceptable for our own internal usage? Or do we need some way to automate flipping that to ECR as part of build?

deployment/scripts/indexd/indexd_setup.sh
```diff
@@ -11,4 +11,4 @@ poetry install -vv --no-dev --no-interaction && poetry show -v
 python3 -m pip install cdispyutils
 popd
 
-/dockerrun.sh
```
Avantol13, line 14: this file isn't used by the Dockerfile, right? b/c above this change this is doing some poetry stuff manually

EliseCastle23: yes! I don't think we need it at all, so I deleted it! Nice catch!

jawadqur (approved):

PR #361, opened 2023-06-05
Change content date defaults to None from ''

k-burt-uch:
### Bug Fixes
- Indexd records no longer return empty strings for content_updated_date and content_updated_time when not provided.

PR #359, opened 2023-05-08
adding labeler to run appro tests

atharvar28:
Jira Ticket: [GQE-74](https://ctds-planx.atlassian.net/browse/GQE-74)
- [ ] Remove this line if you've changed the title to (PXP-xxxx): <title>

### Deployment changes

[GQE-74]: https://ctds-planx.atlassian.net/browse/GQE-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #365, opened 2023-09-01
Fix/content dates for new version

k-burt-uch:
### Bug Fixes
- Add version endpoint now works with content_create_date and content_update_date.


paulineribeyre (dismissed):

.travis.yml
```diff
@@ -9,6 +9,24 @@ sudo: false
 services:
     - postgresql
 
+addons:
+  postgresql: "13"
+  apt:
+    sources:
+      - sourceline: deb http://apt.postgresql.org/pub/repos/apt/ jammy-pgdg main
```
paulineribeyre, line 16: if you have some time to spare, it would be nice to replace this annoying setup with GH actions following @Avantol13's doc [here](https://docs.google.com/document/d/1wf1hQK0hQsVzaK23fIiktrnxYdSgRxOag2VSzUBHFhc) and example [here](https://github.com/uc-cdis/sheepdog/pull/400)

paulineribeyre (dismissed):

BinamB (approved):

.travis.yml
k-burt-uch: Yeah let me knock that out real quick.

k-burt-uch:
Adding @Avantol13 to check the github actions CI

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,21 @@
+name: CI
+on:
+  push:
+    pull_request:
+      types: [opened, reopened]
```
paulineribeyre, line 5:
```suggestion
on:
  push:
  pull_request:
    types: [opened, reopened]
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,21 @@
...
+  Security:
+    name: Security Pipeline
+    uses: uc-cdis/.github/.github/workflows/securitypipeline.yaml@master
+    with:
+       python-poetry: 'false'
+    secrets: inherit
+
+  UnitTest:
+    name: Python Unit Test with Postgres
+    uses: uc-cdis/.github/.github/workflows/python_unit_test.yaml@master
```
paulineribeyre, line 19: the default script runs `poetry run pytest -vv` but in indexd we have coveralls set up and we want to report coverage updates, so i think we actually need a custom script that does this: https://github.com/uc-cdis/indexd/blob/2023.09/.travis.yml#L20-L23

Avantol13: I'd actually be okay leaving this as-is, because if we want to add consistent coverage in the future, we could just modify the template. but we'd love coverage reporting in the meantime. idk, I don't have a strong opinion either way

Avantol13 (approved):

.github/workflows/ci.yaml
k-burt-uch: It seems we've lost coverage reporting already, checking the last few closed PRs for indexd. 

Checking this PR https://github.com/uc-cdis/indexd/pull/361 and this one https://github.com/uc-cdis/indexd/pull/360 there's no coveralls run. It seems the last one was Atharva's PR here: https://github.com/uc-cdis/indexd/pull/359 

I could be missing something, but it seems we've already lost coverage reporting so my vote is to add it to the template in a separate PR. Sound good @paulineribeyre ??

paulineribeyre: I don't know why coveralls is not reporting as a check on those PRs, but you can see in coveralls that it did get the report, for example for #361:
Screenshot 2023-09-28 at 4 00 33 PM

But anyway i'm referring to the coverage badge we have on this repo's readme, more than to the PR checks. It would be nice to include the coveralls logic in the template, but since it's not there yet, i think removing it would be a regression and the coverage % on the repo readme would be inaccurate. Tbh i think it's probably not a heavy lift to add it, but if it is for some reason, we can file a ticket instead

k-burt-uch: Ah, that makes sense. Given that https://github.com/uc-cdis/sheepdog#sheepdog has broken tags now I'll see what I can do about getting this working here and moving it to the template. 

Reading the docs here https://github.com/marketplace/actions/coveralls-github-action it looks fairly straightforward but I might have to DM you with questions if I run into something I can't find in the coveralls config or guides.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,27 @@
...
+    with:
+       python-version: '3.9'
+       test-script: 'tests/ci_commands_script.sh'
+  Coveralls:
```
paulineribeyre, line 23: i don't think this will work because it's set up as a different "job", when it should be a new "step" in the same "job" (to use the report file generated by the `pytest` command). Unfortunately i don't think you can add a step to a job that's defined in a reusable workflow, so if you want to do it that way, it would have to be directly in the reusable workflow.

What i was suggesting was to add the `coveralls` command to `tests/ci_commands_script.sh`. But i think we'd need to pass the token to the `test-script`.

I see now that it's not as straightforward as i thought 😅

paulineribeyre: @k-burt-uch 
https://github.com/uc-cdis/.github/pull/68
https://github.com/uc-cdis/indexd/compare/fix/content-dates-for-new-version...fix/content-dates-for-new-version-pauline

paulineribeyre (approved):

PR #376, opened 2024-02-05
(PPS-107): Feat/single table driver

BinamB:

Coverage went down by 2% because we removed `test_setup.py`. The coverage went down for the `utils.py` file that contains the old, deprecated method of database migration that uses `sqlite3`.

### New Features
- Implemented single table architecture to IndexD with a single new `record` table (Configurable and requires migration)

### Improvements
- Removed `sqlite` from unit tests. Unit tests now only uses `postgres`

### Dependency updates
- `sqlalchemy` ~1.3.3 -> ^1.4.0

### Deployment changes

- sqlite databases are not supported anymore
- To run single table migration check this doc: https://github.com/uc-cdis/indexd/blob/master/docs/migration_to_single_table_indexd.md


Avantol13 (changes requested):
this isn't a full review but figured I share these comments I had so far

Avantol13 (changes requested):
I had some comments before I left for PTO that I forgot to send, apologies. Here they are now

indexd/default_settings.py
```diff
@@ -17,7 +17,7 @@
 # will be created as "<PREFIX><PREFIX><GUID>".
 CONFIG["INDEX"] = {
     "driver": SQLAlchemyIndexDriver(
-        "sqlite:///index.sq3",
+        "postgres://postgres:postgres@localhost:5432/indexd_tests",  # pragma: allowlist secret
```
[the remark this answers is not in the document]
BinamB: Yeah sqlite doesn't handle some of the newer data types. I moved all the test to use postgres instead of sqlite too, just to stay consistent. I saw there was a mix.

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1548 @@
...
+        with self.session as session:
+            for i in session.query(Record):
+                yield i.did
+
+    def totalbytes(self):
+        """
+        Total number of bytes of data represented in the index.
+        """
+        with self.session as session:
+            result = session.execute(select([func.sum(Record.size)])).scalar()
```
Avantol13, line 1239: we could add a simple caching strategy here where we store some global and clear it if records get added. Maybe out of scope for this but worth considering. Same thing for num records

migrations/versions/bb3d7586a096_createsingletable.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "bb3d7586a096"  # pragma: allowlist secret
+down_revision = "a72f117515c5"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
Avantol13, line 20: I think we need another PR after this gets merged to actually clean up the old tables. We can't forget or someone in the future will be super confused.

tests/conftest.py
```diff
@@ -2,22 +2,104 @@
...
 from indexd.auth.errors import AuthError
-from tests import default_test_settings
+from indexd.index.drivers.alchemy import Base as index_base
+from indexd.auth.drivers.alchemy import Base as auth_base
+from indexd.alias.drivers.alchemy import Base as alias_base
+from indexd.index.drivers.alchemy import SQLAlchemyIndexDriver
+from indexd.alias.drivers.alchemy import SQLAlchemyAliasDriver
+from indexd.auth.drivers.alchemy import SQLAlchemyAuthDriver
+
+POSTGRES_CONNECTION = "postgresql://postgres:postgres@localhost:5432/indexd_tests"  # pragma: allowlist secret
```
Avantol13, line 23: isn't there a configuration value for this that the tests use?

tests/conftest.py
```diff
@@ -2,22 +2,104 @@
...
+
+
+def clear_database():
```
Avantol13, line 26: quick docstring pls

tests/conftest.py
```diff
@@ -2,22 +2,104 @@
...
+        # Clear the Auth records
+        auth_driver = SQLAlchemyAuthDriver(POSTGRES_CONNECTION)
+        for model in auth_base.__subclasses__():
+            table = model.__table__
+            delete_statement = table.delete()
+            conn.execute(delete_statement)
+
+
+@pytest.fixture(scope="function", params=["default_settings", "single_table_settings"])
+def combined_default_and_single_table_settings(request):
```
Avantol13, line 69: quick docstring pls

tests/postgres/migrations/test_bb3d7586a096_createsingletable.py
```diff
@@ -0,0 +1,52 @@
+from alembic.config import main as alembic_main
+
+
+def test_upgrade(postgres_driver):
```
Avantol13, line 4: quick docstring pls

tests/postgres/migrations/test_bb3d7586a096_createsingletable.py
```diff
@@ -0,0 +1,52 @@
...
+        ("url_metadata", "jsonb"),
+        ("alias", "ARRAY"),
+    ]
+
+    table_res = conn.execute(get_columns)
+    actual_schema = sorted([i for i in table_res])
+    assert sorted(expected_schema) == actual_schema
+
+
+def test_downgrade(postgres_driver):
```
Avantol13, line 43: quick docstring pls

Avantol13 (changes requested):

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,268 @@
+"""
+
```
Avantol13, line 2: Can you add a quick description here with how to use this?

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,268 @@
...
+        description="Migrate data from old indexd database to new single table database"
+    )
+    parser.add_argument(
+        "creds_path",
+        help="Path to the creds file for the database you're trying to copy data from multi-table to single records table. Defaults to original indexd database creds from the indexd block in the creds.json file.",
+    )
+    args = parser.parse_args()
+    migrator = IndexRecordMigrator(conf_data=args.creds_path)
+    migrator.index_record_to_new_table()
+    # cProfile.run("migrator.index_record_to_new_table()", filename="profile_results.txt")
```
Avantol13, line 265: dead code, remove if not using

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,268 @@
...
+            f"postgresql+psycopg2://{usr}:{psw}@{pghost}:{pgport}/{db}"
+        )
+
+        Base = declarative_base()
+        Base.metadata.create_all(engine)
+        Session = sessionmaker(bind=engine)
+
+        self.session = Session()
+
+    def index_record_to_new_table(self, batch_size=1000, retry_limit=4):
```
Avantol13, line 71: I imagine this batch_size could be much larger. Where's the job that will run this? How much RAM do you plan on providing? Indexd records text is pretty darn small. Maybe 1000 characters on average? 1000 chars / record, 1 char = 1 byte in ASCII, 1kb per record. so 1000 records = 1MB. If we have 1 GBs of usable RAM, we could probably go up to like 10000 records.

Avantol13: I'm not sure we _should_ do that, but if we can reliably, it might be faster (reducing the total number of queries and commits to the db). Would outputting to a file and using /COPY be faster? Don't you guys use /COPY for something already? Did you consider this?

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,268 @@
...
+
+        except Exception as e:
+            self.session.rollback()
+            self.logger.error(f"Errored at {offset}: {e}")
+
+        finally:
+            self.session.close()
+            self.logger.info("Finished migrating :D")
+
+    def get_index_record_hash(self, did):
```
Avantol13, line 148: lets get some docstrings in these functions

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1548 @@
...
+    """
+
+    __tablename__ = "record"
+
+    guid = Column(String, primary_key=True)
+
+    baseid = Column(String, index=True)
+    rev = Column(String)
+    form = Column(String)
+    size = Column(BigInteger, index=True)
```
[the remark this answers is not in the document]
BinamB: removing it, I cant remember why i did that. Its been a while.

bin/migrate_to_single_table.py
BinamB: I did consider this. I think there's a couple of things here:
1. We had a conversation about whether to use sql+bash script vs python. We went with python because it would be easier for any gen3 user to use it. 
2. should we do python + /copy? My concern with this is that this would blow up for something like dcf where we have about 56mil records. I think we'd use up a lot of memory. We could do it in batches though 🤔

migrations/versions/bb3d7586a096_createsingletable.py
BinamB: Yeah definitely. This migration can be an in-between migration where both options are available.

bin/migrate_to_single_table.py
Avantol13: Yeah, it's been a while since we've had the design conversations around the options here. Python makes sense, - we're already so close with this. I was just thinking out loud and not remembering previous conversations. Let's not shift gears at this point. I am curious if you were able to tune these numbers and avoid using OFFSET and if that helped

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+        description="Migrate data from old indexd database to new single table database"
+    )
+    parser.add_argument(
+        "--creds-path",
+        help="Path to the creds file for the database you're trying to copy data from multi-table to single records table. Defaults to original indexd database creds from the indexd block in the creds.json file.",
+    )
+    parser.add_argument(
+        "--start-did",
+        help="did to start at",
+        default=False,
```
Avantol13, line 52: minor thing: I think it's more intuitive if it's `None` instead of `False`

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+        psw = conf_data.get("db_password", "{{db_password}}")
+        pghost = conf_data.get("db_host", "{{db_host}}")
+        pgport = 5432
+
+        try:
+            engine = create_engine(
+                f"postgresql+psycopg2://{usr}:{psw}@{pghost}:{pgport}/{db}"
+            )
+        except Exception as e:
+            print(e)
```
Avantol13, line 78: use logger/logging instead of pure prints

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+                            url_metadata=url_metadata,
+                            acl=acl,
+                            authz=authz,
+                            alias=alias,
+                            record_metadata=metadata,
+                        )
+                    )
+                while records_to_insert:
+                    try:
+                        self.session.bulk_save_objects(records_to_insert)
```
Avantol13, line 149: SQLAlchemy lists this as legacy https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects for 2.0 and beyond. I think since this is a one time run and we're not ported to 2 yet, we're okay

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+                        self.session.commit()
+                        count += len(records_to_insert)
+                        self.logger.info(
+                            f"Done processing {count}/{total_records} records. {(count * 100)/total_records}%"
+                        )
+                        break
+                    except Exception as e:
+                        self.session.rollback()
+                        if "duplicate key value violates unique constraint" in str(e):
```
Avantol13, line 158: instead of this, you should have an except block for the specific Error type that raises that, you can have multiple except blocks in Python so you can split this. out and keep the general rollback case on Exception too

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+                            records_to_insert = self.remove_duplicate_records(
+                                records_to_insert, e
+                            )
+                last_seen_guid = records[-1].did
+        except Exception as e:
+            self.session.rollback()
+            self.logger.error(
+                f"Error in migration: {e}. Last seen guid: {last_seen_guid}. Please "
```
Avantol13, line 166: Please what ? I think the second line didn't get committed

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+                    try:
+                        self.session.bulk_save_objects(records_to_insert)
+                        self.session.commit()
+                        count += len(records_to_insert)
+                        self.logger.info(
+                            f"Done processing {count}/{total_records} records. {(count * 100)/total_records}%"
+                        )
+                        break
+                    except Exception as e:
+                        self.session.rollback()
```
Avantol13, line 157: What's the intended behavior here? It looks like you rollback this batch but then just continue anyway, but you don't have a log for last seen guid. I think you need to do that before exiting. Also I think this should just retry the batch a few times before giving up. 

In fact, this code is pretty nested and getting hard to read. I'd break things out into functions and you can have a process_batch function that you use exponential backoff on

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+        try:
+            stmt = (
+                self.session.query(IndexRecordACE.ace)
+                .filter(IndexRecordACE.did == did)
+                .all()
+            )
+            res = [a.ace for a in stmt]
+            return res
+        except Exception as e:
+            self.logger.error(f"Error with ace for did {did}: {e}")
```
Avantol13, line 248: are you intending to not reraise an error here? so we just skip that field and it gets set as None? I don't think we want that behavior b/c then some records could get "successfully" submitted but if one of these functions failed, that field would be None. This is a good reason to try and avoid early returns b/c that behavior is not obvious. But if you had a guaranteed return value at the end, it would be.

e.g.

```
field_value = None
try:
  get
except: 
  whoops
return field_value
```

That being said, I don't think we want that behavior. I think we want to raise the exception higher and then exponentially backoff trying 1 or 2 more times to get the info for the record.

bin/migrate_to_single_table.py
```diff
@@ -0,0 +1,319 @@
...
+                .all()
+            )
+            res = {key: value for key, value in stmt}
+            return res
+        except Exception as e:
+            self.logger.error(f"Error with alias for did {did}: {e}")
+
+    def remove_duplicate_records(self, records, error):
+        """
+        Remove duplicate records from the bulk insert records list
```
Avantol13, line 304: do you know how this happens in the first place?

tests/conftest.py
```diff
@@ -2,22 +2,110 @@
 import importlib
 import pytest
 import requests
+from sqlalchemy import create_engine
 import mock
 from unittest.mock import patch
 
 # indexd_server and indexd_client is needed as fixtures
-from cdisutilstest.code.indexd_fixture import clear_database
+# from cdisutilstest.code.indexd_fixture import clear_database
```
Avantol13, line 10: dead code

Avantol13 (changes requested):

bin/migrate_to_single_table.py
```diff
@@ -180,6 +158,73 @@ def index_record_to_new_table(self, batch_size=1000):
                 )
             self.logger.info("Finished migrating :D")
 
+    @backoff.on_exception(
+        backoff.expo, Exception, max_tries=5, max_time=60, jitter=backoff.full_jitter
```
Avantol13, line 162: I assume these values (especially `max_time` comes from some testing? It would nice if this script accepted that value as an argument but it's not necessary if you're certain this good after some tests. If this were code we expected to stick around, I'd argue for more config but I think it's okay

bin/migrate_to_single_table.py
```diff
@@ -111,66 +132,23 @@ def index_record_to_new_table(self, batch_size=1000):
...
-                        break
-                    except Exception as e:
-                        self.session.rollback()
-                        if "duplicate key value violates unique constraint" in str(e):
-                            records_to_insert = self.remove_duplicate_records(
-                                records_to_insert, e
-                            )
+                records_to_insert = self.get_info_from_mult_tables(records)
+
+                self.bulk_insert_records(records_to_insert)
```
Avantol13, line 137: How are you handling errors in this new code? It looks like they don't re-raise from this function, so it just continues to run new batches. Is that intended? Are people expected to parse the logs for errors and rerun with a start guid? If so, how do they specify where to "end", e.g. when things started working again. 

The alternative would be for the script to actually die when there's an error by re-raising the exception and providing output so the user knows what batch failed and what the last successful GUID was

Avantol13 (changes requested):

bin/migrate_to_single_table.py
```diff
@@ -132,7 +132,15 @@ def index_record_to_new_table(
...
                     break
 
-                records_to_insert = self.get_info_from_mult_tables(records)
+                try:
+                    records_to_insert = self.get_info_from_mult_tables(records)
+                except Exception as e:
+                    raise Exception(
+                        f"""
+                                    Could not insert records with {e} at offset {offset} with the last seen guid {last_seen_guid}. Please re-run the job with the following command
+                                    gen3 job run indexd-single-table-migration-job --start-did {last_seen_guid}
```
Avantol13, line 141: The service isn't fully aware of the job, so I'd leave the specific details out. I'd just say "Please re-run the job with the following `start-did`: {last_seen_guid}"

Avantol13: shouldn't this error handling also be around the bulk_insert_records?

Avantol13 (dismissed):

bin/migrate_to_single_table.py
Avantol13: Yeah but these aren't catching the exception just because the fields are [], right? Like urls can be empty, but `get_urls_record` should still return []. I'm talking about the case where there really is an exception and you catch it.

The db query fails b/c we lose connection to the db. So right now you log that, but you don't return anything or reraise the exception, so the code continues on. I don't _think_ we want that behavior. If the db connection is broken, we want to raise an exception. And higher up, if we get an exception, we want to retry getting the record a few times just in case it was something flaky.

bin/migrate_to_single_table.py
Avantol13: I'd say we remove this then, it's probably adding a decent chunk of time iterating through all the records

pyproject.toml
```diff
@@ -12,6 +12,7 @@ include = [
 [tool.poetry.dependencies]
```
Avantol13, line 12: need a version bump above

Avantol13:
At this point, most of my major feedback has been addressed. I'm dismissing my blocking review requesting changes so that progress can be made if I don't get to this before I head out on PTO.

Things I was not able to complete:

- Manual testing, running the service the old way, migrating, then running the new way

Things that need to be finished before a final approval:

- Passing integration tests
- Ideally another dev manually testing

indexd/single_table_settings.py
```diff
@@ -0,0 +1,69 @@
+import os
+
+from indexd.index.drivers.single_table_alchemy import SingleTableSQLAlchemyIndexDriver
```
paulineribeyre, line 3: Most of `single_table_settings` is a copy/merge of `default_settings` and `default_test_settings` - I notice a few changes but not sure why they're needed, for example `DIST` is different and `DRS_SERVICE_INFO.id` is not in all the settings files

Is adding a new file necessary? Can't we switch between `SQLAlchemyIndexDriver` and `SingleTableSQLAlchemyIndexDriver` depending on another config field, or even an env var?

If that's not possible, can we remove the duplicated settings and just overwrite `default_settings` with whatever is in `single_table_settings`?
Maybe `single_table_settings` could do `from indexd.default_settings import *` like `default_test_settings` does, and just set a different `CONFIG["INDEX"]`

pyproject.toml
```diff
@@ -12,6 +12,7 @@ include = [
...
 python = ">=3.9,<4.0"
 alembic = "^1.9.4"
+asyncpg = "^0.29.0"
```
paulineribeyre, line 15: you mentioned this can be removed now

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+                        match_string = "$.* ? ({})".format(matches)
+                        query = query.filter(
+                            func.jsonb_path_exists(Record.url_metadata, match_string)
+                        )
+
+            if negate_params:
+                query = self._negate_filter(session, query, **negate_params)
+
+            if page is not None:
+                query = query.order_by(Record.updated_date)
```
paulineribeyre, line 215: there was a comment in the multi-table driver that i think we should keep:
```
if page is not None:
    # order by updated date so newly added stuff is
    # at the end (reduce risk that a new records ends up in a page
    # earlier on) and allows for some logic to check for newly added records
    # (e.g. parallelly processing from beginning -> middle and ending -> middle
    #       and as a final step, checking the "ending"+1 to see if there are
    #       new records).
    query = query.order_by(IndexRecord.updated_date)
else:
    query = query.order_by(IndexRecord.did)
```

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+            record.guid = guid
+            record.baseid = baseid
+            record.rev = str(uuid.uuid4())[:8]
+            record.form = form
+            record.size = size
+            record.file_name = file_name
+            record.version = version
+            record.description = description
+            record.content_created_date = content_created_date
+            record.content_updated_date = content_updated_date
```
paulineribeyre, line 985: in the multi-table code, these dates were validated by `_validate_and_format_content_dates`, do we not need to validate here anymore?

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+                # create/update for that resource (old authz and new authz)
+                old_authz = [u for u in record.authz] if record.authz else []
+                all_authz = old_authz + authz
+                try:
+                    auth.authorize("update", all_authz)
+                    authorized = True
+                except AuthError as err:
+                    self.logger.error(
+                        authz_err_msg.format("update", all_authz)
+                        + " Falling back to 'file_uplaod' on '/data_file'."
```
paulineribeyre, line 571:
```suggestion
                        + " Falling back to 'file_upload' on '/data_file'."
```

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+                auth.authorize("delete", resources)
+            except AuthError as err:
+                self.logger.warning(
+                    f"Auth error while deleting all aliases for did {did}: User not authorized to delete one or more of these resources: {resources}"
+                )
+                raise err
+
+            query = session.query(Record).filter(Record.guid == did)
+            record = query.one()
+            # delete this GUID's aliases and add new aliases
```
paulineribeyre, line 729:
```suggestion
            # delete this GUID's aliases
```

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+            result.append(resp_dict)
+        return result
+
+
+def check_url_metadata(url_metadata, record):
+    """
+    create url metadata record in database
+    """
+    urls = {u for u in record.urls}
+    for url, metadata in url_metadata.items():
```
paulineribeyre, line 1507:
```suggestion
    for url in url_metadata:
```

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+            record.version = version
+            record.description = description
+            record.content_created_date = content_created_date
+            record.content_updated_date = content_updated_date
+            record.urls = urls
+            record.acl = acl
+            record.authz = authz
+            record.hashes = hashes
+            record.record_metadata = metadata
+            record.url_metadata = check_url_metadata(urls_metadata, record)
```
paulineribeyre, line 991: i don't think `check_url_metadata()` needs to return anything, there are no changes to the input `urls_metadata` - the only benefit i see is being able to write this one-liner 😄

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+            new_record = Record()
+            guid = new_did
+            if not guid:
+                guid = str(uuid.uuid4())
+                if self.config.get("PEPREND_PREFIX"):
+                    guid = self.config["DEFAULT_PREFIX"] + guid
+
+            new_record.guid = guid
+            new_record.baseid = old_record.baseid
+            new_record.rev = str(uuid.uuid4())
```
paulineribeyre, line 1048:
```suggestion
            new_record.rev = str(uuid.uuid4())[:8]
```

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1522 @@
...
+def get_record_if_exists(did, session):
+    """
+    Searches for a record with this did and returns it.
+    If no record found, returns None.
+    """
+    return session.query(Record).filter(Record.guid == did).first()
+
+
+SCHEMA_MIGRATION_FUNCTIONS = []
+CURRENT_SCHEMA_VERSION = len(SCHEMA_MIGRATION_FUNCTIONS)
```
paulineribeyre, line 1522: does something break if you remove these?

migrations/versions/bb3d7586a096_createsingletable.py
paulineribeyre: Can we add a TODO in this file?

tests/conftest.py
```diff
@@ -2,22 +2,109 @@
 import importlib
 import pytest
 import requests
+from sqlalchemy import create_engine
 import mock
 from unittest.mock import patch
 
 # indexd_server and indexd_client is needed as fixtures
-from cdisutilstest.code.indexd_fixture import clear_database
```
paulineribeyre, line 9: i think that was the last use of it, so we should be able to remove the `cdisutilstest` dependency! 🎉

deployment/Secrets/indexd_settings.py
```diff
@@ -1,10 +1,9 @@
 from os import environ
 import json
 import config_helper
-from indexd.index.drivers.alchemy import SQLAlchemyIndexDriver
 from indexd.alias.drivers.alchemy import SQLAlchemyAliasDriver
 from indexd.auth.drivers.alchemy import SQLAlchemyAuthDriver
-
+from indexd.index.drivers.single_table_alchemy import SingleTableSQLAlchemyIndexDriver
```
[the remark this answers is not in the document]
BinamB: We change the `CONFIG["INDEX"]["driver"]` from `SQLAlchemyIndexDriver` to `SingleTableSQLAlchemyIndexDriver`
Should i do it a different way? Some logic to make it simple?

tests/conftest.py
```diff
@@ -2,22 +2,109 @@
...
+        # Clear the Auth records
+        auth_driver = SQLAlchemyAuthDriver(POSTGRES_CONNECTION)
+        for model in auth_base.__subclasses__():
+            table = model.__table__
+            delete_statement = table.delete()
+            conn.execute(delete_statement)
+
+
+@pytest.fixture(scope="function", params=["default_settings", "single_table_settings"])
+def combined_default_and_single_table_settings(request):
```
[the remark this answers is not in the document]
BinamB: There are tests that shouldn't use the fixture

.secrets.baseline
```diff
@@ -239,7 +239,7 @@
         "filename": "tests/default_test_settings.py",
```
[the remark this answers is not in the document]
paulineribeyre: Can you mention the migration in "Deployment changes", say it's optional and how to run it

deployment/Secrets/indexd_settings.py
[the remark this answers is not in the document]
paulineribeyre: i think a config flag would nice, something like
```
# one of: "multitable" (default), "single_table"
DATABASE_FORMAT = ""
```
and then initializing `CONFIG["INDEX"]` with
```
if DATABASE_FORMAT.lower() == "single_table":
    CONFIG["INDEX"] = ...
else:
    CONFIG["INDEX"] = ...
```
but it depends on what we want to do about [this](https://github.com/uc-cdis/indexd/pull/376#discussion_r1747361996)

tests/conftest.py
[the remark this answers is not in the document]
paulineribeyre: for tests that shouldn't use the fixture, shouldn't we use the single table by default instead of the multitable? [here](https://github.com/uc-cdis/indexd/blob/bad7c84aa036d2471bb9be99a5c049521be595ab/tests/conftest.py#L110-L115)
since single table will eventually be the default

tests/conftest.py
```diff
@@ -2,22 +2,109 @@
...
+        importlib.reload(single_table_settings)
+        single_table_settings.settings = {
+            **single_table_settings.settings,
+        }
+        yield get_app(single_table_settings.settings)
+
+    try:
+        clear_database()
+    except Exception:
+        pass
```
paulineribeyre, line 101: we might want to log the exception instead of silently passing (same in `app` and `user`)

tests/conftest.py
```diff
@@ -29,18 +116,31 @@ def app():
...
+
     yield {
         "Authorization": ("Basic " + base64.b64encode(b"test:test").decode("ascii")),
         "Content-Type": "application/json",
     }
-    app.auth.delete("test")
+
+    try:
+        driver.delete("test")
```
paulineribeyre, line 139: the `add` and `delete` functions still exist in the auth driver, why does this need to change?

tests/test_driver_alchemy_crud.py
```diff
@@ -17,78 +17,76 @@
...
-    with sqlite3.connect("index.sq3") as conn:
-        driver = SQLAlchemyIndexDriver("sqlite:///index.sq3")
 
-        count = conn.execute(
-            """
-            SELECT COUNT(*) FROM index_record
-        """
-        ).fetchone()[0]
+    engine = create_engine(POSTGRES_CONNECTION)
+    driver = SQLAlchemyIndexDriver(POSTGRES_CONNECTION)
```
paulineribeyre, line 32: i'm confused why you write `driver = SQLAlchemyIndexDriver(POSTGRES_CONNECTION)` in a bunch of tests that use `combined_default_and_single_table_settings`. When we use the single table it shouldn't be that driver(?)

there are 2 `test_driver_init_does_not_create_records` tests, the other one doesn't include `driver = SQLAlchemyIndexDriver(POSTGRES_CONNECTION)`, is it needed here?

tests/test_driver_alchemy_crud.py
```diff
@@ -845,69 +852,72 @@ def test_driver_get_all_versions_with_no_record():
...
+def test_driver_update_record(skip_authz, combined_default_and_single_table_settings):
     _test_driver_update_record()
 
 
-@util.removes("index.sq3")
 def _test_driver_update_record():
     """
     Tests updating of a record.
     """
-    with sqlite3.connect("index.sq3") as conn:
```
paulineribeyre, line 894: "Deployment changes" should mention that sqlite databases are not supported anymore

tests/test_migration.py
```diff
@@ -0,0 +1,148 @@
+import random
```
paulineribeyre, line 1: could we rename this file `test_migrate_to_single_table`? and if possible, move it to `postgres/migrations`?

indexd/index/drivers/single_table_alchemy.py
BinamB: Changed it. Running check_url_metadata first and then setting the url_metadata outside of the function

tests/conftest.py
BinamB: This ones a little complicated. Its the way tests are set. I did it a while ago and idk why it doesnt work but when i remove the driver, the tests start to fail

tests/test_migration.py
BinamB: I changed the name of the file but moving it to the `postgres/migrations` directory makes the second test get stuck at migration for some reason

tests/test_driver_alchemy_crud.py
BinamB: Went ahead and cleaned this up. I removed the `combined_default_and_single_table_settings` since these are tests that test direct queries to the old tables

indexd/index/drivers/single_table_alchemy.py
BinamB: removed

indexd/single_table_settings.py
BinamB: Removing the new file and consolidating everything in conftest

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1539 @@
...
+                session.add(record)
+                session.commit()
+            except IntegrityError:
+                raise MultipleRecordsFound("{guid} already exists".format(guid=guid))
+
+            self._validate_and_format_content_dates(
+                record=record,
+                content_created_date=content_created_date,
+                content_updated_date=content_updated_date,
+            )
```
paulineribeyre, line 1019: `_validate_and_format_content_dates` should probably be called before `session.add` and `session.commit`

docs/migration_to_single_table_indexd.md
```diff
@@ -0,0 +1,103 @@
...
+    gen3 psql indexd -c 'create database indexd_new'
+    # dump old db and restore it on the new one
+    gen3 db backup indexd | psql -U $indexd_user -h <hostname> -d indexd_new
+```
+
+**If you don’t have permissions:**
+
+a. Run `gen3 db creds indexd`
+
+b. Using the information from above, run `gen3 psql $g3Farmserver -c "alter user $username createdb;"`
```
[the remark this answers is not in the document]
BinamB: Yep! I worked with PE to come of with this

indexd/index/drivers/single_table_alchemy.py
BinamB: oops

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -0,0 +1,1539 @@
...
+            record.hashes = hashes
+            record.record_metadata = metadata
+
+            self._validate_and_format_content_dates(
+                record=record,
+                content_created_date=content_created_date,
+                content_updated_date=content_updated_date,
+            )
+            record.content_created_date = content_created_date
+            record.content_updated_date = content_updated_date
```
paulineribeyre, line 1010:
```suggestion
```
it looks like `_validate_and_format_content_dates` sets the right values, so we shouldn't overwrite them

paulineribeyre (approved):
🎉 🚀

PR #383, opened 2024-10-01
Fix delete endpoints in Single Table

BinamB:

### Bug Fixes
- Fix resource authorize checks for delete and alias endpoints in single table alchemy


MaribelleHGomez (approved):
lgtm

PR #386, opened 2025-02-05
Add Indexes to Single Table

BinamB:

### Improvements
- Add indexes to `record` table

### Deployment changes

- Requires alembic migration to `b61e2d4aa93d`


indexd/index/drivers/single_table_alchemy.py
```diff
@@ -51,22 +51,22 @@ class Record(Base):
...
     rev = Column(String)
     form = Column(String)
-    size = Column(BigInteger)
+    size = Column(BigInteger, index=True)
     created_date = Column(DateTime, default=datetime.datetime.utcnow)
     updated_date = Column(DateTime, default=datetime.datetime.utcnow)
-    file_name = Column(String)
-    version = Column(String)
-    uploader = Column(String)
+    file_name = Column(String, index=True)
```
Avantol13, line 57: I don't think we should be this generous with indexing. There are drawbacks on insertion and other considerations. I don't think you should index on anything other than `size`, `hashes`. guid and baseid should already be indexed.

What prompted the addition of all of these?

Avantol13 (changes requested):

PR #391, opened 2025-03-10
Fix version endpoint, update deps

paulineribeyre:

### Bug Fixes
- Fix the `_version` endpoint not to return empty values

### Dependency updates
- Update dependencies´´


Dockerfile
```diff
@@ -14,24 +14,27 @@ FROM base AS builder
...
-    && VERSION=`git describe --always --tags` && echo "VERSION=\"${VERSION}\"" >> /${appname}/version_data.py
+RUN git config --global --add safe.directory ${appname} && COMMIT=`git rev-parse HEAD` && echo "COMMIT=\"${COMMIT}\"" > ${appname}/version_data.py \
+    && VERSION=`git describe --always --tags` && echo "VERSION=\"${VERSION}\"" >> ${appname}/version_data.py
 
 # Final stage
 FROM base
 
 COPY --from=builder /${appname} /${appname}
 
+RUN dnf -y install vim
```
nss10, line 36: I think having vim and nano in the base image would still be a better idea. Can approve this for now though!

nss10 (approved):

paulineribeyre:
ignoring known audit service test failure

PR #405, opened 2025-08-04
feature/rbac-2

bwalsh:
### New Features

- Introduced Role-Based Access Control (RBAC) to indexd.
  - Added support for enforcing authorization on database operations via Arborist (`adds RBAC to db operations`).
  - Integrated a cached call to Arborist to reduce authorization lookup overhead (`add cached call to Arborist`).
  - Added configuration flag to enable or disable RBAC enforcement (`Adds RBAC config`).

### Breaking Changes

- None. The RBAC feature is gated by a configuration flag and maintains backward compatibility when disabled.

### Bug Fixes

- Improved error messaging for unauthorized access and token validation (`Improve error handling`).

### Improvements

- Added developer documentation for RBAC configuration and usage (`developer documentation`).
- Added test coverage for RBAC behavior (`test RBAC`).

### Dependency updates

- None.

### Deployment changes

- New optional configuration setting: `RBAC`. When set to `True`, RBAC enforcement is active for protected records.
- Arborist must be reachable by the indexd service for RBAC to function properly.

### DIFF
--- a/indexd/auth/__init__.py
+++ b/indexd/auth/__init__.py
@@ -3,7 +3,8 @@ from functools import wraps
 from flask import current_app
 from flask import request
 
-from .errors import AuthError
+from indexd.auth.errors import AuthError, AuthzError
+from indexd.errors import UserError
 
 
 def authorize(*p):
@@ -20,8 +21,8 @@ def authorize(*p):
 
         @wraps(f)
         def check_auth(*args, **kwargs):
-            if not request.authorization:
-                raise AuthError("Username / password required.")
+            if not request.authorization.parameters.get("username"):
+                raise AuthError(f"Basic auth Username / password required. {request.authorization}")
             current_app.auth.auth(
                 request.authorization.parameters.get("username"),
                 request.authorization.parameters.get("password"),
@@ -31,13 +32,20 @@ def authorize(*p):
 
         return check_auth
     else:
-        method, resources = p
+        method, resources_ = p
         if request.authorization and request.authorization.type == "basic":
             current_app.auth.auth(
                 request.authorization.parameters.get("username"),
                 request.authorization.parameters.get("password"),
             )
         else:
-            if not isinstance(resources, list):
-                raise UserError(f"'authz' must be a list, received '{resources}'.")
-            current_app.auth.authz(method, list(set(resources)))
+            if not isinstance(resources_, list):
+                raise UserError(f"'authz' must be a list, received '{resources_}'.")
+            return current_app.auth.authz(method, list(set(resources_)))
+
+
+def resources():
+    """
+    Returns a list of resources the user has access to. Uses Arborist if available.
+    """
+    return current_app.auth.resources()
--- a/indexd/auth/drivers/__init__.py
+++ b/indexd/auth/drivers/__init__.py
@@ -1,0 +1,24 @@
+import functools
+import time
+
+
+def timed_cache(ttl_seconds):
+    """
+    Decorator to cache the result of a function for a specified time-to-live (TTL) in seconds.
+    """
+    def decorator(func):
+        cache = {}
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            key = functools._make_key(args, kwargs, typed=False)
+            now = time.time()
+            if key in cache:
+                result, timestamp = cache[key]
+                if now - timestamp < ttl_seconds:
+                    return result
+            result = func(*args, **kwargs)
+            cache[key] = (result, now)
+            return result
+        return wrapper
+    return decorator
--- a/indexd/auth/drivers/alchemy.py
+++ b/indexd/auth/drivers/alchemy.py
@@ -1,4 +1,5 @@
 import hashlib
+import sys
 
 from contextlib import contextmanager
 
@@ -11,6 +12,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.ext.declarative import declarative_base
 
 from indexd.auth.driver import AuthDriverABC
+from indexd.auth.drivers import timed_cache
 
 from indexd.auth.errors import AuthError, AuthzError
 
@@ -133,14 +135,13 @@ class SQLAlchemyAuthDriver(AuthDriverABC):
         try:
             # A successful call from arborist returns a bool, else returns ArboristError
             try:
-                authorized = self.arborist.auth_request(
-                    get_jwt_token(), "indexd", method, resource
-                )
+                authorized = self.cached_auth_request(get_jwt_token(), "indexd", method, resource)
             except Exception as e:
                 logger.error(
                     f"Request to Arborist failed; now checking admin access. Details:\n{e}"
                 )
                 authorized = False
+
             if not authorized:
                 # admins can perform all operations
                 is_admin = self.arborist.auth_request(
@@ -157,7 +158,53 @@ class SQLAlchemyAuthDriver(AuthDriverABC):
                             "The indexd admin '/programs' logic is deprecated. Please update your policy to '/services/indexd/admin'"
                         )
                 if not is_admin:
-                    raise AuthError("Permission denied.")
+                    raise AuthError("Permission denied. (not is_admin)")
         except Exception as err:
             logger.error(err)
             raise AuthzError(err)
+
+    def resources(self):
+        """
+        Returns a list of resources for the given user.
+        """
+        if not self.arborist:
+            raise AuthError(
+                "Arborist is not configured; cannot perform authorization check"
+            )
+        token = get_jwt_token()
+        try:
+            resources = self.caching_auth_mapping(token)
+            return resources
+        except Exception as err:
+            raise AuthError(
+                "Failed to get resources from Arborist. Please check your Arborist configuration."
+            )
+
+    @timed_cache(1800)  # Cache for 30 minutes (typical JWT expiration time)
+    def caching_auth_mapping(self, token):
+        """
+        Returns a list of resources the user has access to.
+        Uses Arborist if available.
+        If a token is provided, it will use that token to get the auth mapping.
+        If no token is provided, it will use the default auth mapping.
+        """
+        if token:
+            resources = self.arborist.auth_mapping(
+                jwt=token
+            )
+        else:
+            resources = self.arborist.auth_mapping()
+        return resources
+
+    @timed_cache(1800)  # Cache for 30 minutes (typical JWT expiration time)
+    def cached_auth_request(self, token, service, method, resource):
+        """
+        Makes an authenticated request to Arborist and caches the result.
+        This method is used to check if the user has access to a specific resource
+        with a specific method.
+        """
+        return self.arborist.auth_request(
+            token, service, method, resource
+        )
+
+
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -1,5 +1,6 @@
 import os
-import re
+
+import cdislogging
 import flask
 import json
 from indexd.errors import AuthError, AuthzError
@@ -7,6 +8,8 @@ from indexd.errors import UserError
 from indexd.index.errors import NoRecordFound as IndexNoRecordFound
 from indexd.errors import IndexdUnexpectedError
 from indexd.utils import reverse_url
+
+logger = cdislogging.get_logger(__name__)
 
 blueprint = flask.Blueprint("drs", __name__)
 
@@ -367,6 +370,15 @@ def handle_no_index_record_error(err):
 def handle_unexpected_error(err):
     ret = {"msg": err.message, "status_code": err.code}
     return flask.jsonify(ret), err.code
+
+
+@blueprint.errorhandler(Exception)
+def handle_uncaught_exception(err):
+    """
+    Handle uncaught exceptions and log them.
+    """
+    logger.error(err, exc_info=True)
+    return flask.jsonify(error=f"Internal server error"), 500
 
 
 @blueprint.record
--- a/indexd/index/drivers/alchemy.py
+++ b/indexd/index/drivers/alchemy.py
@@ -2,6 +2,8 @@ import datetime
 import uuid
 import json
 from contextlib import contextmanager
+
+import flask
 from cdislogging import get_logger
 from sqlalchemy import (
     BigInteger,
@@ -24,6 +26,7 @@ from sqlalchemy.orm import joinedload, relationship, sessionmaker
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 from indexd import auth
+from indexd.auth.errors import AuthzError
 from indexd.errors import UserError, AuthError
 from indexd.index.driver import IndexDriverABC
 from indexd.index.errors import (
@@ -35,6 +38,79 @@ from indexd.index.errors import (
 from indexd.utils import migrate_database
 
 Base = declarative_base()
+
+
+def _get_authorized_resources(requested_method="READ") -> list[str]:
+    """Retrieve the user's authorized resources from Arborist."""
+    # Token is __not__ required for RBAC, but we can use it to get the user's context.
+    # See https://github.com/uc-cdis/indexd/pull/400#discussion_r2243298256
+    # `Arborist allows no token to be sent on purpose, it allows assignment of anonymous access. So we don't want to raise this error here`
+    # if not flask.request.authorization:
+    #     raise AuthError("Authorization header is required for RBAC")
+    # token = flask.request.headers.get("Authorization")
+    # if token is None:
+    #     raise AuthError("Authorization header is required for RBAC")
+    authorized_resources = []
+    for resource, permissions in auth.resources().items():
+        for permission in permissions:
+            method = permission['method']
+            service = permission['service']
+            if requested_method == "READ":
+                if "read-storage" == method or service == "indexd":
+                    authorized_resources.append(resource)
+            else:
+                if method in ["create", "update", "delete"]:
+                    authorized_resources.append(resource)
+    return authorized_resources
+
+
+def _check_user_access(authorized_resources: list[str], authz: list[str]) -> None:
+    """Raise an AuthzError if the user is not authorized for the given resources."""
+    for resource in authz:
+        if resource not in authorized_resources:
+            raise AuthzError(
+                f"User is not authorized for project {resource} {authz} {type(authz)}"
+            )
+
+
+def _is_rbac() -> bool:
+    """ Check if RBAC is enabled for the current app."""
+    # For simplicity, we check the config directly.
+    return flask.current_app.config.get('RBAC', False)
+
+
+def _enforce_rbac(authz, method="READ") -> tuple[list[str], list[str]]:
+    """ Enforce RBAC for the current request."""
+    any_authz = None
+    if _is_rbac():
+        # If RBAC is not enabled, we don't need to check user access.
+        # This is for backwards compatibility with existing code.
+        authorized_resources = _get_authorized_resources(method)
+        if len(authorized_resources) == 0:
+            # Force filtering by an non-existent resource if no resources are authorized.
+            authorized_resources = ["NO-AUTHORIZED-RESOURCES"]
+        if authz:
+            _check_user_access(authorized_resources, authz or [])
+            any_authz = None
+        else:
+            authz = None
+            any_authz = authorized_resources
+
+    return any_authz, authz
+
+
+def _enforce_record_authz(record):
+    """ Enforce record authorization based on the current request's RBAC settings."""
+    if not _is_rbac():
+        return
+    authorized_resources = _get_authorized_resources()
+    if len(authorized_resources) == 0:
+        raise AuthError("User is not authorized for any resources")
+    if isinstance(record, IndexRecord):
+        record = record.to_document_dict()
+    if not isinstance(record, dict):
+        raise UserError("Record must be a dictionary")
+    _check_user_access(authorized_resources, record.get("authz", []))
 
 
 class BaseVersion(Base):
@@ -389,11 +465,14 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
         metadata=None,
         ids=None,
         urls_metadata=None,
-        negate_params=None,
+        negate_params=None
     ):
         """
         Returns list of records stored by the backend.
         """
+
+        any_authz, authz = _enforce_rbac(authz)
+
         with self.session as session:
             query = session.query(IndexRecord)
 
@@ -450,7 +529,12 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
                     query = query.filter(IndexRecord.did.in_(sub.subquery()))
             elif authz == []:
                 query = query.filter(IndexRecord.authz == None)
-
+            elif any_authz:
+                # if any_authz is set, we want to filter records that have ANY of the authz elements
+                sub = session.query(IndexRecordAuthz.did).filter(
+                    IndexRecordAuthz.resource.in_(any_authz)
+                )
+                query = query.filter(IndexRecord.did.in_(sub.subquery().select()))
             if hashes:
                 for h, v in hashes.items():
                     sub = session.query(IndexRecordHash.did)
@@ -534,6 +618,7 @@ class SQLAlchemyIndexDriver(IndexDriverABC):
                 query = query.offset(limit * page)
 
             return [i.to_document_dict() for i in query]
+
 
     @staticmethod
     def _negate_filter(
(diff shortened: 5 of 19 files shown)


bwalsh:
Not yet addressed:
- [ ] Ensure ARE_RECORDS_DISCOVERABLE, GLOBAL_DISCOVERY_AUTHZ See [discussion](https://github.com/uc-cdis/indexd/pull/400#discussion_r2243579240)
- [ ] Add a corresponding feature flag to helm chart

@Avantol13 - could you review and comment
---

## 🧾 User Story 1: Control Whether Records Are Discoverable

**Title**: Configurable Discovery of Indexd Records

**As a** platform operator,
**I want** to control whether indexd records are discoverable at all via a config flag,
**So that** I can prevent users from listing or retrieving records unless explicitly permitted.

### Acceptance Criteria

* [ ] Given `ARE_RECORDS_DISCOVERABLE=False`, when a client sends a request (with valid token):
  * indexd returns a `403 Forbidden` for all reads with id e.g. `GET /index/<did>` 
  * indexd filters out all records without read permission e.g. `GET /index` 
* [ ] Given `ARE_RECORDS_DISCOVERABLE=True`, the RBAC rules are ignored 
* [ ] This behavior is documented in `indexd_settings.py` and the README with a description of impact on runtime behavior.

---

## 🧾 User Story 2: Global Discovery Authorization Control

**Title**: Global Discovery Authz for Indexd Records, Support Discovery Access Independent from File Access

**As a** system administrator,
**I want** to configure a global authorization group for reading/discovering indexd records,
**So that** discovery can be gated separately from file access and we can support user registration workflows.

**As a** data commons architect,
**I want** to decouple discovery access (e.g. listing/searching records) from access to the underlying files,
**So that** I can implement workflows like "register to see what’s available", then "apply for access to download".

### Acceptance Criteria

Assuming ARE_RECORDS_DISCOVERABLE=False

* [ ] If `GLOBAL_DISCOVERY_AUTHZ=None`, then RBAC will use record-level `authz` fields are used to authorize `GET` requests to records.  ie then record-level `authz` continues to govern access to records.
* [ ] If `GLOBAL_DISCOVERY_AUTHZ is set` **and** if a user has permissions to the resource set in GLOBAL_DISCOVERY_AUTHZ, then RBAC will ignore filters for record-level `authz` fields and return all records.
* [ ] Behavior is clearly documented, including the override effect of `GLOBAL_DISCOVERY_AUTHZ`.

---

## 📌 Configuration Summary

```python
# Whether any records are discoverable at all
ARE_RECORDS_DISCOVERABLE = True  # default: True

# Override per-record authz for GET/read
# Only applies to record discovery (not file access)
# If None, use per-record `authz`
GLOBAL_DISCOVERY_AUTHZ = ["/indexd/discovery"]
```

Avantol13:
In general the comments above look good, thanks for all the detail.

This part:

>  indexd returns a 403 Forbidden for all reads with id e.g. GET /index/<did>

I think needs to actually behave similar to READ filtering based on config. In other words, if you request a did and you do have access to authz, this should return 200. If you request a did and do you have access to the global authz that's configured, this should return 200. Basically this should only 403 in situations where this record itself would've been filtered out.

indexd/auth/drivers/alchemy.py
```diff
@@ -157,7 +158,53 @@ def authz(self, method, resource):
...
+        """
+        if token:
+            resources = self.arborist.auth_mapping(
+                jwt=token
+            )
+        else:
+            resources = self.arborist.auth_mapping()
+        return resources
+
+    @timed_cache(1800)  # Cache for 30 minutes (typical JWT expiration time)
```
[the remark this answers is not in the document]
bwalsh: @Avantol13 

Re. [requirements](https://github.com/uc-cdis/indexd/pull/400#discussion_r2243324670)


> has to be dynamic based on the expiration of the token

Understood.  At the same time, [previous feedback](https://github.com/uc-cdis/indexd/pull/400#discussion_r2243298256)  stated:

> Arborist allows no token to be sent on purpose, it allows assignment of anonymous access.  

Additionally, AFAIK, no validation of the token occurs now in indexd. ie no calls to `authutils.token.validate_jwt()`

So, if there is a token:
* [x]  🆕  we can check to ensure it has not expired, use expiry time as ttl
* [x] already being used as a cache key0

If there is no token:
* [x] 🆕  use maximum_ttl_seconds as ttl
* [x] 🆕  add authentication header to cache key (for basic and no auth)

Other:
* [x] 🆕  clean up any unused cache entries

indexd/drs/blueprint.py
```diff
@@ -369,6 +372,15 @@ def handle_unexpected_error(err):
     return flask.jsonify(ret), err.code
 
 
+@blueprint.errorhandler(Exception)
+def handle_uncaught_exception(err):
```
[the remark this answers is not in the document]
bwalsh: done

bwalsh:
> In general the comments above look good, thanks for all the detail.
> 
> This part:
> 
> > indexd returns a 403 Forbidden for all reads with id e.g. GET /index/
> 
> I think needs to actually behave similar to READ filtering based on config. In other words, if you request a did and you do have access to authz, this should return 200. If you request a did and do you have access to the global authz that's configured, this should return 200. Basically this should only 403 in situations where this record itself would've been filtered out.

Thanks.  I edited the comment above

bwalsh:
On reviews complete:
- [ ] squash commits

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+                    authorized_resources.append(resource)
+    return authorized_resources
+
+
+def _check_user_access(authorized_resources: list[str], authz: list[str]) -> None:
+    """Raise an AuthzError if the user is not authorized for the given resources."""
+    for resource in authz:
+        if resource not in authorized_resources:
+            raise AuthzError(
+                f"User is not authorized for project {resource} {authz} {type(authz)}"
```
Avantol13, line 72: see other comments, let's put the details here into a debug log and leave the authzerror opaque. `User is not authorized`

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+            )
+
+
+def _is_rbac() -> bool:
```
Avantol13, line 76: see other comment, let's remove all "RBAC" and replace with something like `_is_record_discovery_enabled`

indexd/index/drivers/alchemy.py
```diff
@@ -720,7 +828,7 @@ def add(
         hashes = hashes or {}
         metadata = metadata or {}
         urls_metadata = urls_metadata or {}
-
+        _enforce_rbac(authz, method="WRITE")
```
Avantol13, line 831: can we rely on existing authz for write and not overload this feature with that scope? I'd rather keep the record management separate from the external-facing record discovery ability

indexd/index/drivers/query/urls.py
```diff
@@ -44,6 +44,8 @@ def query_urls(
             versioned.lower() in ["true", "t", "yes", "y"] if versioned else None
         )
 
+        any_authz, authz = _enforce_rbac([])
```
Avantol13, line 47: we need a better name for all these (the two variables and the function) - I can't tell what their intention is.

is this something like:

`user_authz_resources, ?? what is authz`? and why are we passing an empty list?

Let's logically separate these maybe and refactor into something to get the authz resources if you need them (which it looks like you do).

indexd/index/drivers/query/urls.py
```diff
@@ -52,12 +54,14 @@ def query_urls(
                 IndexRecordUrl.did, q_func["string_agg"](IndexRecordUrl.url, ",")
             )
 
+            # Only join IndexRecord once, using outerjoin if versioned filtering is needed
+            if versioned is not None or any_authz:
```
Avantol13, line 58: why did we need to add `or any_authz` ?

indexd/index/drivers/query/urls.py
Avantol13: and let's not do any of this unless we need to (e.g. record discovery is enabled) and let's make that clear at the high-level here. 

I would move all this down to the section where you have `if any_authz` and instead the if condition should be `if ARE_RECORDS_DISCOVERABLE is False` and then do all the new logic

indexd/index/drivers/query/urls.py
```diff
@@ -80,13 +84,37 @@ def query_urls(
...
+
             # [('did', 'urls')]
+
             record_list = (
                 query.order_by(IndexRecordUrl.did.asc())
                 .offset(offset)
                 .limit(limit)
                 .all()
             )
+            # alternatively, we could have used the following query:
```
Avantol13, line 110: let's just remove this unless there's a reason to switch to it in the future

indexd/index/drivers/query/urls.py
Avantol13: actually.. the driver should not be relying on a stateful request at all. We need to refactor the portion where we obtain user authz out of this and pass it in as `filter_authz_resources` in here. That'd actually be much cleaner - no config checks at all in here, just add the filter if you need to, calling code determines if there needs to be a filter at all.

indexd/index/drivers/alchemy.py
Avantol13: actually I'm confused on the terminology and overall function here. Is this a global config check or a check on a specific user? We need to refactor all this a bit to make the whole process more clear.

Roughly the logic should be:

1. Request comes in.

2. If `ARE_RECORDS_DISCOVERABLE` is True, we should fallback to existing behavior. Do this check as high level as possible. It should be clear reading through the code where this feature flag turns on and off behavior.

3. if `ARE_RECORDS_DISCOVERABLE` is False, we need to check if `GLOBAL_DISCOVERY_AUTHZ` and then continue to authz check to see if user has that. 

4. if no `GLOBAL_DISCOVERY_AUTHZ`, then continue to authz check to see if user has access to records authz. **OR if this is a list endpoint**: Get all user authz, pass that as filter into driver to only return those records. 

The fact that seemingly 2 and part of 3 are mingled in this one function is a little confusing

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+
+def _enforce_rbac(authz, method="READ") -> tuple[list[str], list[str]]:
+    """ Enforce RBAC for the current request."""
+    any_authz = None
+    if _is_rbac():
+        # If RBAC is not enabled, we don't need to check user access.
+        # This is for backwards compatibility with existing code.
+        authorized_resources = _get_authorized_resources(method)
+        if len(authorized_resources) == 0:
+            # Force filtering by an non-existent resource if no resources are authorized.
```
Avantol13, line 105: why is this required?

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+    return any_authz, authz
+
+
+def _enforce_record_authz(record):
+    """ Enforce record authorization based on the current request's RBAC settings."""
+    if not _is_rbac():
+        return
+    authorized_resources = _get_authorized_resources()
+    if len(authorized_resources) == 0:
+        raise AuthError("User is not authorized for any resources")
```
Avantol13, line 123: see other comments, leave this vague, just say user is not authorized

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+    if isinstance(record, IndexRecord):
+        record = record.to_document_dict()
+    if not isinstance(record, dict):
+        raise UserError("Record must be a dictionary")
```
Avantol13, line 127: I don't think this is a user error. Can this happen in normal runtime or are you trying to catch a perceived edge case? either way, this should not be a UserError unless it's the user's "fault" by something they passed in

indexd/auth/drivers/alchemy.py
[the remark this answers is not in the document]
bwalsh: done

indexd/index/drivers/query/urls.py
bwalsh: > actually.. the driver should not be relying on a stateful request at all.  

I had a version that did this - checked config in the calling code.  Unfortunately, as far as I can see, that means adding the code to all the blueprints both to check config and change calls.

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+    authorized_resources: list = _get_authorized_resources()
+    # if any of the global discovery authz resources are in the authorized resources, RBAC is not enabled
+    if any(resource in authorized_resources for resource in global_discovery_authz):
+        return False
+    # If we reach here, RBAC is enabled.
+    return True
+
+
+def _enforce_rbac(authz, method="READ") -> tuple[list[str], list[str]]:
+    """ Enforce RBAC for the current request."""
```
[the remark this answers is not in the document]
bwalsh: done

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
 Base = declarative_base()
 
 
+def _get_authorized_resources(requested_method="READ") -> list[str]:
+    """Retrieve the user's authorized resources from Arborist."""
+    # Token is __not__ required for RBAC, but we can use it to get the user's context.
+    # See https://github.com/uc-cdis/indexd/pull/400#discussion_r2243298256
+    # `Arborist allows no token to be sent on purpose, it allows assignment of anonymous access. So we don't want to raise this error here`
```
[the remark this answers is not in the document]
bwalsh: done

indexd/auth/__init__.py
```diff
@@ -31,13 +32,20 @@ def check_auth(*args, **kwargs):
...
         else:
-            if not isinstance(resources, list):
-                raise UserError(f"'authz' must be a list, received '{resources}'.")
-            current_app.auth.authz(method, list(set(resources)))
+            if not isinstance(resources_, list):
+                raise UserError(f"'authz' must be a list, received '{resources_}'.")
+            return current_app.auth.authz(method, list(set(resources_)))
+
+
+def resources():
```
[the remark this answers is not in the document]
bwalsh: done

indexd/index/drivers/alchemy.py
```diff
@@ -37,6 +40,94 @@
...
+    #     raise AuthError("Authorization header is required for RBAC")
+    # token = flask.request.headers.get("Authorization")
+    # if token is None:
+    #     raise AuthError("Authorization header is required for RBAC")
+    authorized_resources = []
+    for resource, permissions in auth.resources().items():
+        for permission in permissions:
+            method = permission['method']
+            service = permission['service']
+            if requested_method == "READ":
```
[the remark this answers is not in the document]
bwalsh: done

indexd/index/drivers/query/urls.py
bwalsh: done

indexd/index/drivers/alchemy.py
bwalsh: done

indexd/index/drivers/alchemy.py
bwalsh: done

indexd/index/drivers/query/urls.py
bwalsh: done

indexd/index/drivers/query/urls.py
bwalsh: other points except dependency on stateful request - done

indexd/index/drivers/alchemy.py
bwalsh: done

Avantol13 (changes requested):
re: db driver and how to centrally organize things. Here's my current thinking:

We should, theoretically, be able to move all the new code we need with stateful decisions *out* of the db driver b/c nothing really needs the db.

Here's my idea:

Put the authz check for discovery in a similar authorize decorator to this https://github.com/uc-cdis/indexd/blob/b5b198a9bf42f5b61042975f4d123f951a981bf7/indexd/auth/__init__.py#L9, maybe call it `authorize_discovery` and add that decorator everywhere you need. The logic in there should look like this:

* if is_discovery_enabled (check config only)
    * Get user's authz (perhaps cache, could even use flask's per-request cache flask.g if that somehow simplifies - I know that won't save beyond the request)
    * if config GLOBAL_AUTHZ set
         * Check if the GLOBAL_authz is in the user's authz
    * if config GLOBAL not set
         * Check if user's authz contains records authz (this will require making a db call based on the request's ID)

done. Now we have appropriately denied access pre-blueprint logic with this decorator.

Within the blueprints that need the logic for filtering, now we can implement a shared set of util functions.

Before making a db query:

     * if is_discovery_enabled (check config only)
          * get user authz (perhaps from cache)
          * don't worry about GLOBAL AUTHZ at all b/c we already handled that in the new decorator of this endpoint, so we only get here if they are authorized
          * Add filter for records to db query based on user authz

This way, the stateful logic stays out of the db driver and in the request handling (which is where it should be) and we have minimized duplication of code as much as possible.

What do you think?

indexd/index/drivers/query/urls.py
```diff
@@ -4,7 +4,7 @@
 from indexd.index.drivers.alchemy import (
     IndexRecord,
     IndexRecordUrl,
-    IndexRecordUrlMetadata,
+    IndexRecordUrlMetadata, _filter_authorized_resources, IndexRecordAuthz, _is_record_discovery_enabled,
```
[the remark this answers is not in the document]
bwalsh: done

indexd/index/drivers/query/urls.py
```diff
@@ -80,13 +82,28 @@ def query_urls(
...
                 )
-            # [('did', 'urls')]
+
+            # add authz filter
+            if _is_record_discovery_enabled():
+                permitted_authz_resources, authz = _filter_authorized_resources()
+                # if permitted_authz_resources is set, we want to filter records that have ANY of the authz elements
+                # was getting the following warning
+                #  SAWarning: SELECT statement has a cartesian product between FROM element(s) "index_record" and FROM element "index_record_url".  Apply join condition(s) between each element to resolve. .all()
+                # This warning means the query is selecting from multiple tables (index_record and index_record_url) without a proper join condition, causing a cartesian product. To fix this, so we explicitly join IndexRecordUrl to IndexRecord using their relationship.
```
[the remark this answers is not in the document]
bwalsh: done

indexd/auth/drivers/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+def request_auth_cache():
+    """
+    Decorator to cache the result of a function for a specified maximum TTL in seconds.
+    The actual cache duration is determined by the 'token' parameter's expiration.
+    If no token is provided as a argument to the cached token, use the Authentication header's token and included the Authorization header is in the cache key.
+    """
+
+    def decorator(func):
+        cache = {}
```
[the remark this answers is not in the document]
bwalsh: added a comment in the code back to this description

indexd/auth/drivers/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+            # Clean up any old cache entries
+            if not token_ttl:
+                token_ttl = 0
+            keys_to_delete = [k for k, (v, t) in cache.items() if now - t >= token_ttl]
+            for k in keys_to_delete:
+                del cache[k]
+
+            return result
+
+        def calculate_ttl(now, token) -> int | None:
```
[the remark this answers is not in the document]
Avantol13: don't calculate ttl, just put the token and exp in the cache and instead of calculating it every time you check the cache, just get the entries with exp that isn't past now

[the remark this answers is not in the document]
Avantol13: it would simplify a lot of this code imo

bwalsh:
## Clarification question:

How will a web tier decorator will retrieve the `record's authz`?  There are several possible database calls, each with their own set of possible parameters.  Unclear how to a decorator would manage this.  Assuming it could, wouldn't that double the db IO count for successful queries (once in the decorator + once in the driver)?

### from above 

image

Avantol13:
> ## Clarification question:
> 
> How will a web tier decorator will retrieve the `record's authz`? There are several possible database calls, each with their own set of possible parameters. Unclear how to a decorator would manage this. Assuming it could, wouldn't that double the db IO count for successful queries (once in the decorator + once in the driver)?
> ### from above
> image

The decorator would need to use the driver to interact as the data access layer (DAL) based on the content of the request. e.g. take GUID, ask DAL for record, get GUID's authz from record. In the backwards-compatible case (where discovery is open) then all this logic should get skipped. In the new case, this does mean that you db hit to get the authz.

I'm open to reconsidering the decorator as the approach (this was just a cursory idea) so long as we figure out a way that the separation b/t web / DAL exists appropriately, but note that the decorator is just a function wrapping the decorated function, so you can pass in variables. e.g. theoretically you could get the index record from the db in the decorator logic (to get authz) and then pass the whole record into the endpoint function to avoid a double db hit (and in the case where we don't use this feature, just pass none and then hit the db if you don't have the record when you need it).

bwalsh:
Thanks for the thoughtful review, @Avantol13 I’ve aligned the behavior and added a small infrastructure layer to make it reliable and testable.

### What changed (referencing commit `2064504`)

* **Discovery context + decorator.** Introduced a `ContextVar`-backed `discovery_context` and an `@authorize_discovery` decorator that injects `can_user_discover` and `authorized_resources` into the driver entry points. This keeps the repo/DB code Flask-agnostic and avoids depending on `flask.request` while still giving us request-scoped auth state. See commit **“adds authorize_discovery decorator”** (`206450476c81bec42cff222702093c2a3583eb50`). ([[GitHub](https://github.com/uc-cdis/indexd/pull/405/commits/206450476c81bec42cff222702093c2a3583eb50)])

* **Plain-parameter injection.** The **affected driver functions** (e.g., `ids(...)` see above) were updated to accept **plain old parameters** `can_user_discover` and `authorized_resources` which the decorator injects when not explicitly provided by the caller.

* This commit includes `indexd/auth/discovery_context.py` and related wiring plus unit tests to verify the injection semantics. ([[GitHub](https://github.com/uc-cdis/indexd/pull/405/commits/206450476c81bec42cff222702093c2a3583eb50)])


### Why this addresses the concern

The updated logic and the discovery context make that true for both list/search and direct-DID paths, while keeping the enforcement code decoupled from Flask and easy to unit test. ([[GitHub](https://github.com/uc-cdis/indexd/pull/405)])

*Impact on blueprints & compatibility*

* No blueprint rework required. The change is wired at the app edge (before_request seeding the context) and at the repository boundary via the decorator. Blueprint routes/endpoints remain untouched.

* Preserves existing functionality. This is additive and non-breaking; no API or schema changes.

* Existing tests continue to pass. The current test suite runs without modification; new unit tests cover the decorator/context behavior.


### Follow-ups

* I’ll open a small PR in **gen3-helm** to surface `ARE_RECORDS_DISCOVERABLE` and `GLOBAL_DISCOVERY_AUTHZ` as explicit Helm values and update docs accordingly (not done yet; tracked as a follow-up from the earlier discussion). ([[GitHub](https://github.com/uc-cdis/indexd/pull/405)][2])

If anything looks off in the commit, I’m happy to tweak. Thanks again for the clear guidance, this made the behavior much more predictable.

indexd/auth/discovery_context.py
```diff
@@ -0,0 +1,127 @@
...
+    # Remove global discovery authz resources from authorized_resources to avoid confusion
+    authorized_resources = [_ for _ in authorized_resources if _ not in global_discovery_authz]
+
+    # if are_records_discoverable is True, then can_user_discover is True for everyone
+    if are_records_discoverable:
+        can_user_discover = True
+
+    set_auth_context(are_records_discoverable=are_records_discoverable,
+                     can_user_discover=can_user_discover,
+                     authorized_resources=authorized_resources)
```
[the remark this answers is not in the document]
bwalsh: Agreed, web tier will resolve config and authorized_resources on demand via `auth_context()` and pass to db layer.  See latest change to `get_all_index_record_versions`:

> Updates the get_all_index_record_versions view to call auth_context() and pass can_user_discover and authorized_resources to the index driver’s get_all_versions() method.

indexd/auth/discovery_context.py
```diff
@@ -0,0 +1,127 @@
...
+import inspect
+from functools import wraps
+from typing import Any, Mapping
+
+
+import flask
+
+from indexd import auth
+
+discovery_context = ContextVar("discovery_context", default={})
```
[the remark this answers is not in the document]
bwalsh: Agreed, ensure_auth_context will no longer be called.  Note that a "context" var  holder is no longer as we are passing the variables explicitly

>> The main thing I'm trying to ensure is appropriate separation between the web and data layer. We don't want explicit or implicit coupling. The data layer should not need to understand web layer. The endpoint handling the incoming request should be translating what it can and interacting with data through an explicit, well-defined interface, not relying on an implicit context being available

PR #433, opened 2026-03-25
Indexd static options

nrandorf:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/BDC-975

### New Features
- Added static OPTIONS endpoint for single DRS object and associated pytests

### DIFF
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -85,8 +85,9 @@ def get_drs_object_options(object_id):
     # Get static authz metadata
     authz_metadata = blueprint.drs_authorization_metadata
 
+    # If static match detected, update with object id
     if authz in authz_metadata.keys():
-        print("- authz match detected")
+        authz_metadata[authz].update({"drs_object_id": object_id})
         return flask.jsonify(authz_metadata[authz]), 200
     return flask.jsonify(data), 200
 
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -76,20 +76,26 @@ def get_drs_object_options(object_id):
     @blueprint.route("/ga4gh/drs/v1/objects/<path:object_id>", methods=["OPTIONS"]) #test b: same URL
     """
     print("- inside options... ")
-    # Extract authz based on guid
-    expand = True if flask.request.args.get("expand") == "true" else False
-    ret = blueprint.index_driver.get_with_nonstrict_prefix(object_id)
-    authz = ret["authz"][0]
-    print(f"authz: {authz}")
+    # Get authz based on guid
+    try:
+        ret = blueprint.index_driver.get_with_nonstrict_prefix(object_id)
+        authz = ret["authz"][0]
+    except IndexNoRecordFound as err:
+        return handle_no_index_record_error(err)
 
-    # Get static authz metadata
-    authz_metadata = blueprint.drs_authorization_metadata
+    # Get static authz metadata based on blueprint
+    try:
+        # Get static authz metadata
+        authz_metadata = blueprint.drs_authorization_metadata
+        # If static match detected, update with object id
+        if authz in authz_metadata.keys():
+            authz_metadata[authz].update({"drs_object_id": object_id})
+            return flask.jsonify(authz_metadata[authz]), 200
+        return flask.jsonify(data), 200
 
-    # If static match detected, update with object id
-    if authz in authz_metadata.keys():
-        authz_metadata[authz].update({"drs_object_id": object_id})
-        return flask.jsonify(authz_metadata[authz]), 200
-    return flask.jsonify(data), 200
+    # Otherwise return unexpected error
+    except IndexdUnexpectedError as err:
+        return handle_unexpected_error(err)
 
 
 @blueprint.route("/ga4gh/drs/v1/objects", methods=["GET"])
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -66,6 +66,31 @@ def get_drs_object(object_id):
     return flask.jsonify(data), 200
 
 
+@blueprint.route(
+    "/ga4gh/drs/v1/options/objects/<path:object_id>", methods=["OPTIONS"]
+)  # test a (add 'options' to URL)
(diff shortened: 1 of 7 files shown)


indexd/drs/blueprint.py
```diff
@@ -64,6 +68,33 @@ def get_drs_object(object_id):
...
+
+    # Get static authz metadata based on blueprint
+    try:
+        # Get static authz metadata
+        authz_metadata = blueprint.drs_authorization_metadata
+        # If static match detected, update with object id
+        if authz in authz_metadata.keys():
+            authz_metadata[authz].update({"drs_object_id": object_id})
+            return flask.jsonify(authz_metadata[authz]), 200
+        return flask.jsonify(data), 200
```
SpencerAxelrod, line 94: where does data come from?

indexd/drs/blueprint.py
```diff
@@ -64,6 +68,33 @@ def get_drs_object(object_id):
...
+    except IndexNoRecordFound as err:
+        return handle_no_index_record_error(err)
+
+    # Get static authz metadata based on blueprint
+    try:
+        # Get static authz metadata
+        authz_metadata = blueprint.drs_authorization_metadata
+        # If static match detected, update with object id
+        if authz in authz_metadata.keys():
+            authz_metadata[authz].update({"drs_object_id": object_id})
```
SpencerAxelrod, line 92: have to be careful here. Updating blueprint.drs_authorization_metadata updates the shared dict with the **last-requested object ID**. So this could be a race condition if another request comes in before we finish. Make authz_metadata a copy of it instead of a reference I think

indexd/drs/blueprint.py
```diff
@@ -64,6 +68,33 @@ def get_drs_object(object_id):
...
+            return flask.jsonify(authz_metadata[authz]), 200
+        return flask.jsonify(data), 200
+
+    # Otherwise return unexpected error
+    except IndexdUnexpectedError as err:
```
SpencerAxelrod, line 97: does anything in the try raise an IndexdUnexpectedError? I think test_auth_options_unexpected_error passes coincidentally because NameError produces a 500 from flask. And Namerror come from the ^data which is undefined?

indexd/drs/blueprint.py
```diff
@@ -64,7 +69,36 @@ def get_drs_object(object_id):
...
+    except IndexNoRecordFound as err:
+        return handle_no_index_record_error(err)
+
+    # Get static authz metadata based on blueprint
+    try:
+        # Get static authz metadata
+        authz_metadata = copy.deepcopy(blueprint.drs_authorization_metadata)
+        # Return unexpected error if static match not detected
+        if authz not in authz_metadata.keys():
+            return handle_unexpected_error(IndexdUnexpectedError)
```
SpencerAxelrod: You may need to instantiate an error instance here, something like:

`handle_unexpected_error(IndexdUnexpectedError(code=500, message="No authorization metadata configured for authz path"))`

Or can we just maybe `raise IndexdUnexpectedError`  here? As it will be be then sent to handle just a fiew lines below this

indexd/drs/blueprint.py
```diff
@@ -103,6 +137,84 @@ def list_drs_records():
...
+                {"error_code": 500, "object_ids": unexpected_error_guids}
+            )
+        # Update compiled results
+        compiled_info = {}
+        compiled_info["summary"] = summary
+        compiled_info["unresolved_drs_objects"] = unresolved_drs_objects
+        compiled_info["resolved_drs_objects"] = resolved_drs_objects
+
+    # Otherwise major unexpected error encountered, return defaults
+    except IndexdUnexpectedError as err:
```
SpencerAxelrod: Same thing here, I'm not sure anything in the `try` raises an IndexdUnexpetedError

indexd/drs/blueprint.py
```diff
@@ -103,6 +137,84 @@ def list_drs_records():
...
+                if authz not in authz_metadata.keys():
+                    unexpected_error_guids.append(doc["did"])
+                    continue
+                # otherwise update with object id and save info for return
+                authz_metadata[authz].update({"drs_object_id": doc["did"]})
+                resolved_drs_objects.append(authz_metadata[authz])
+                resolved_count = resolved_count + 1
+
+            #  Record unexpected errors & continue to next
+            except:
```
SpencerAxelrod: I think this should at least be `except Exception as err` so we can at least log what happened. Also just `except:` will catch all possible errors, some things where we likely want to raise

indexd/bulk/blueprint.py
```diff
@@ -24,25 +25,7 @@ def bulk_get_documents():
...
-        )
-        query = query.options(joinedload(IndexRecord.acl))
-        query = query.options(joinedload(IndexRecord.authz))
-        query = query.options(joinedload(IndexRecord.hashes))
-        query = query.options(joinedload(IndexRecord.index_metadata))
-        query = query.options(joinedload(IndexRecord.aliases))
-        query = query.filter(IndexRecord.did.in_(ids))
-
-    docs = [q.to_document_dict() for q in query]
+    docs = blueprint.index_driver.get_bulk(ids)
```
SpencerAxelrod: I think we are losing performance. Note the previous implementation used eager loading using `.joinedload`. 

Basically with sqlalchemy, without the joined loads it queries the db in `.to_document_dict()` (e.g. when in calls `self.urls`, `self.acl`, etc.) for each attribute you need (lazy loading). Which means we are querying the db it multiple times for a single record. The eager loading queries it all upfront more efficiently so `.to_document_dict()` isn't doing all the querying.

This should only affect multi-table indexd tho

indexd/index/drivers/single_table_alchemy.py
```diff
@@ -807,6 +807,16 @@ def get(self, guid, expand=True):
 
             return record.to_document_dict()
 
+    def get_bulk(self, guid_list, expand=True):
```
SpencerAxelrod: why do we have `expand=true` here in single table but not in multi table? The 2 should maybe have same params, and maybe add `get_bulk` to the abstract class `IndexDriverABC`?

PR #378, opened 2024-05-28
Feat/stats table

rishabh3456:
Link to JIRA ticket if there is one: 

### New Features

Added a stats table to track number of files stored and total bytes used by the system monthly for faster "/_stats" endpoint response.

Changes are applied to both single-table and multi-table schemas. All changes were made on the feat/single-table-driver branch.

PR #435, opened 2026-03-26
Feat/stats

SpencerAxelrod:

Updating #378

[Link to feature doc](https://docs.google.com/document/d/1emYw1laZ-YGN2ixMg7M54RDb45QRfrROVEWOvsTcT1I)

### New Features
- Pre-computed stats table replaces full-table `COUNT(*)/SUM(size)` on GET /_stats
- Historical stats queries: GET /_stats?month=X&year=Y

### Deployment changes

- reconcile_stats.py can be used in the container for stats reconciliation. Could be a Cronjob?


indexd/index/drivers/alchemy.py
```diff
@@ -1794,6 +1861,34 @@ def delete_bundle(self, bundle_id):
...
+                            StatsRecord.year < int(year),
+                        )
+                    )
+                    .order_by(StatsRecord.year.desc(), StatsRecord.month.desc())
+                    .first()
+                )
+                return (stats.total_record_count, stats.total_record_bytes)
+            except Exception as e:
+                self.logger.warning(f"Failed to get stats: {e}")
+                return (None, None)
```
paulineribeyre: Let's raise an exception instead of returning None. Actually, let's just remove the whole try/catch block, there is no reason we should expect this to fail

paulineribeyre (dismissed):

indexd/index/drivers/alchemy.py
```diff
@@ -1794,6 +1861,34 @@ def delete_bundle(self, bundle_id):
...
+                stats = (
+                    session.query(StatsRecord)
+                    .filter(
+                        or_(
+                            and_(
+                                StatsRecord.month <= int(month),
+                                StatsRecord.year == int(year),
+                            ),
+                            StatsRecord.year < int(year),
+                        )
```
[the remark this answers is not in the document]
SpencerAxelrod: we need for example: if you query for january 2026, it would exclude december 2025 (which might be the most recent)

indexd/index/drivers/alchemy.py
SpencerAxelrod: Yes that makes sense, I will remove try catch. But we would still need some kind of none check and return (0, 0), for legitimate historical queries from before the first row right? Like if first record is 2025, and they query for 2020?

paulineribeyre: True

indexd/index/drivers/alchemy.py
```diff
@@ -382,6 +376,44 @@ def update_stats(session, number, size=None):
...
+                    StatsRecord.year == int(year),
+                ),
+                StatsRecord.year < int(year),
+            )
+        )
+        .order_by(StatsRecord.year.desc(), StatsRecord.month.desc())
+        .first()
+    )
+    if stats is None:
+        return (0, 0)
```
paulineribeyre, line 422: could you add a quick comment to explain when we might get this edge case

paulineribeyre: and unit test

indexd/index/drivers/alchemy.py
```diff
@@ -783,6 +870,7 @@ def add(
 
                 if self.config.get("ADD_PREFIX_ALIAS"):
                     self.add_prefix_alias(record, session)
+                update_stats(session, 1, size or 0)
```
paulineribeyre: oh now i understand why you were doing `if size is None: size = 0` in `update_stats` 🤦‍♀️ Sorry, the previous version was better, can you put it back instead of passing `size or 0` as argument

indexd/stats_table_migration.py
```diff
@@ -7,32 +7,29 @@
...
+import sqlalchemy as sa
 from cdislogging import get_logger
 from sqlalchemy import and_, func
 
+from indexd.index.drivers.alchemy import IndexRecord, StatsRecord
+
 logger = get_logger(__name__)
 
 
 def seed_stats_from_connection(bind):
```
paulineribeyre, line 19: The file could be renamed `stats_utils.py` or the functions could be moved to `indexd/utils.py`

indexd/index/drivers/alchemy.py
paulineribeyre: Could you add a unit test for what you mentioned above:
> we would still need some kind of none check and return (0, 0), for legitimate historical queries from before the first row right? Like if first record is 2025, and they query for 2020?

indexd/stats_table_migration.py
```diff
@@ -1,93 +0,0 @@
-"""
```
paulineribeyre: you deleted this file but did not commit `stats_utils.py`

migrations/versions/9a2169051163_createstatstable.py
```diff
@@ -0,0 +1,41 @@
...
+        ),
+        sa.Column(
+            "total_record_bytes", sa.BIGINT(), autoincrement=False, nullable=False
+        ),
+        sa.Column("month", sa.INTEGER(), autoincrement=False, nullable=False),
+        sa.Column("year", sa.INTEGER(), autoincrement=False, nullable=False),
+    )
+    op.create_index("ix_stats_year_month", "stats", ["year", "month"])
+
+    from indexd.stats_table_migration import seed_stats_from_connection
```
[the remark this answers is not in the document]
paulineribeyre: ⬆️

indexd/index/drivers/alchemy.py
paulineribeyre: ⬆️

indexd/index/drivers/alchemy.py
```diff
@@ -327,6 +339,89 @@ def get_record_if_exists(did, session):
...
+                    StatsRecord.year == int(year),
+                ),
+                StatsRecord.year < int(year),
+            )
+        )
+        .order_by(StatsRecord.year.desc(), StatsRecord.month.desc())
+        .first()
+    )
+    if stats is None:
+        # This is the case where stats row exists.
```
paulineribeyre:
```suggestion
        # This is the case where no stats row exists.
```

indexd/index/drivers/alchemy.py
```diff
@@ -327,6 +339,89 @@ def get_record_if_exists(did, session):
...
+        # NOTE: The Alembic migration seeds the stats table automatically on first deploy,
+        # so this should be rare (e.g. a manual intervention).
+        # We COULD fall back to a COUNT(*)/SUM(size) on index_record (expensive)
+        # In this case fix the root cause by running the reconciliation command,
+        # which recomputes stats from index_record and backfills the stats table:
+        #   python bin/reconcile_stats.py
+        return (0, 0)
```
paulineribeyre, line 422: isn't this ⬇️  the edge case?
> legitimate historical queries from before the first row right? Like if first record is 2025, and they query for 2020?

paulineribeyre (approved):

PR #441, opened 2026-04-30
Feat/drs 1.5

k-burt-uch:

### New Features
- New endpoint: `OPTIONS /objects/{object_id}` -  returns a list of Authorizations that can be used to determine how to authorize requests to get object or post object.
- New endpoint: `OPTIONS /objects` - Returns authorization information for multiple DrsObjects at once.
- New endpoint: `POST /objects` - Get info about multiple DrsObjects
- New endpoint: `POST /objects/access (Bulk Access URLs)` - Returns an array of URL objects that can be used to download of multiple DrsObjects.
- Update to DrsObject's `access_methods`:  `region`, `cloud`, `available` and `authorization` are added to fields. 
### Deployment changes
- CLOUD_PROVIDER_MAP and DRS_AUTHORIZATION_METADATA environment variables should be provided to support the cloud field for access_methods and the authorizations field for access_methods. See docs for more information.


indexd/drs/blueprint.py
```diff
@@ -217,13 +415,16 @@ def indexd_to_drs(record, expand=False):
             location_type = location.split(":")[
                 0
             ]  # (s3, gs, ftp, gsiftp, globus, htsget, https, file)
+            cloud = blueprint.cloud_provider_map.get(location_type)
```
SpencerAxelrod: Per the feature doc we need to support http mappings e.g. 

```
CLOUD_PROVIDER_MAP = {
    "s3": "aws",
    "gs": "gcp",
    "az": "azure",
    "https": {
        "m3.aicommons.com/ai": "aws",
        "storage.googleapis.com": "gcp",
    },
}
```

For an HTTPS URL, location_type is "https" and if CLOUD_PROVIDER_MAP contains a nested dict under "https", cloud becomes that nested dict object. We probably need a helper function here and in the non drs indexd called like `derive_cloud(url)` that handles both simple scheme mappings and nested HTTPS prefix mappings

k-burt-uch: Do we have cases for this in gen3-code-vigil?

SpencerAxelrod: no it doesn't. And tbh it may be out of scope in the very short term

SpencerAxelrod:
are we adding the bulk drs object endpoint in this PR? `POST /ga4gh/drs/v1/objects`

indexd/drs/blueprint.py
k-burt-uch: Let me add the cases here. We can backlog a ticket on GQE.

k-burt-uch: http mappings added.

k-burt-uch:
> are we adding the bulk drs object endpoint in this PR? `POST /ga4gh/drs/v1/objects`

Yes - this fell through the tracks on tasking 🤦 

I'll push an update later today

indexd/drs/blueprint.py
```diff
@@ -64,7 +105,34 @@ def get_drs_object(object_id):
...
+    # Get static authz metadata based on blueprint
+    try:
+        # Get static authz metadata
+        authz_metadata = copy.deepcopy(blueprint.drs_authorization_metadata)
+        # Otherwise, match exists & we can updated with object id
+        authz_metadata[authz].update({"drs_object_id": object_id})
+        return flask.jsonify(authz_metadata[authz]), 200
+    # Otherwise catch unknown error
+    except Exception as err:
+        return handle_unexpected_error(err)
```
SpencerAxelrod: looking at the `handle_unexpected_error()` definition, it only is registered to catch `IndexdUnexpectedError`. If `authz_metadata[authz]` gets a KeyError, it will be passed to `handle_unexpected_error` and crash. Tests don't catch this cuz it will cause a 500 but not for why we expect. 

We could think about:

1. Specifically create an `IndexdUnexpectedError`

```
except Exception as err:
    return handle_unexpected_error(IndexdUnexpectedError(code=500, message=str(err)))
```

2. Maybe just add another except e.g.:

```
except KeyError:
    return flask.jsonify({"msg": f"No authorization metadata configured for {authz}", "status_code": 404}), 404
```

indexd/drs/blueprint.py
```diff
@@ -64,7 +105,34 @@ def get_drs_object(object_id):
...
-@blueprint.route("/ga4gh/drs/v1/objects", methods=["GET"])
+@blueprint.route("/ga4gh/drs/v1/objects/<path:object_id>", methods=["OPTIONS"])
+def get_drs_object_options(object_id):
+    """
+    Returns a specific DRSobject metadata with object_id
+    """
+    # Get authz based on guid
+    try:
+        ret = blueprint.index_driver.get_with_nonstrict_prefix(object_id)
+        authz = ret["authz"][0]
```
SpencerAxelrod: do we have a case here where `ret["authz"]` would be an empty list? like an open/public DRS object?

indexd/drs/blueprint.py
```diff
@@ -1,43 +1,56 @@
...
+
+logger = get_logger(__name__)
 
 blueprint = flask.Blueprint("drs", __name__)
 
 blueprint.config = dict()
 blueprint.index_driver = None
 blueprint.service_info = {}
+blueprint.cloud_provider_map = {}
+blueprint.max_bulk_request_length = 100
```
SpencerAxelrod, line 20: I think we should also add `blueprint.drs_authorization_metadata = {}` here in case a commons does not have `DRS_AUTHORIZATION_METADATA` in config

indexd/drs/blueprint.py
SpencerAxelrod: also, is this where we need to support multple authz resources? Will `ret["authz"]` contain more than one resource path here?

indexd/drs/blueprint.py
```diff
@@ -47,10 +60,38 @@ def get_drs_service_info():
...
     return flask.jsonify(ret), 200
 
 
-@blueprint.route("/ga4gh/drs/v1/objects/<path:object_id>", methods=["GET"])
+@blueprint.route(
+    "/ga4gh/drs/v1/objects/<path:object_id>",
+    methods=["GET"],
+    provide_automatic_options=False,
+)
 def get_drs_object(object_id):
```
SpencerAxelrod, line 90: Don't we need to add authorizations to the access methods for `GET /objects` and `GET /objects/{id}`?

indexd/drs/blueprint.py
```diff
@@ -103,6 +171,201 @@ def list_drs_records():
...
+                drs_object = indexd_to_drs(doc, expand=True)
+
+                authz = doc["authz"][0]
+                authz_metadata = copy.deepcopy(blueprint.drs_authorization_metadata)
+
+                if authz not in authz_metadata:
+                    unexpected_error_guids.append(did)
+                    continue
+
+                authorization_info = authz_metadata[authz]
```
SpencerAxelrod: I had imagined authorizations to be resolved in `indexd_to_drs`, looks like here we are adding them here in the route handler. If we resolve instead in `index_to_drs`, then we are also adding them for `GET /objects` and `GET /objects/{id}`

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+            "passport_auth_issuers": [],
+        }
+
+        authz_path_list_length = len(authz_path_list)
+        # If index driver found no object auth path info, return empty authz data
+        if authz_path_list_length == 0:
+            return compiled_metadata_details
+        # If auth path is for open project, just return default auth info
+        # Note: if multiple paths exists and one is an open project, only default info is gserviceaccount
+        if any(["/programs/open_access/projects" in path for path in authz_path_list]):
```
k-burt-uch:
```suggestion
        if any(["/open" in path for path in authz_path_list]):
```

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+            authz_metadata_details = authz_metadata[authz]
+            # Reduce issuer list(s) to a sorted set to address potential duplicates..
+            if "passport_auth_issuers" in authz_metadata_details:
+                compiled_passport_auth_issuers = sorted(
+                    set(authz_metadata_details["passport_auth_issuers"]).union(
+                        set(compiled_passport_auth_issuers)
+                    )
+                )
+            if "bearer_auth_issuers" in authz_metadata_details:
+                compiled_bearer_auth_issuers = sorted(
```
k-burt-uch: Why are we sorting the resulting set?

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+            return compiled_metadata_details
+        # If auth path is for open project, just return default auth info
+        # Note: if multiple paths exists and one is an open project, only default info is gserviceaccount
+        if any(["/programs/open_access/projects" in path for path in authz_path_list]):
+            compiled_metadata_details["supported_types"] = ["BearerAuth"]
+            compiled_metadata_details["bearer_auth_issuers"] = [
+                blueprint.default_bearer_issuer
+            ]
+            return compiled_metadata_details
+        # If auth path doesn't match config auth data, return empty authz data
```
k-burt-uch: We should return the default bearer issuer. If there's a default passport issuer configured, that should be returned too. 

```
        compiled_metadata_details = {
            "drs_object_id": object_id,
            "supported_types": [BearerToken],
            "bearer_auth_issuers": [<default_bearer>],
            "passport_auth_issuers": [],
        }
```
And if default passport issuer has a truthy value:

```
        compiled_metadata_details = {
            "drs_object_id": object_id,
            "supported_types": [BearerToken, Passport],
            "bearer_auth_issuers": [<default_bearer>],
            "passport_auth_issuers": [<passport issuer],
        }
```

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+            if path not in config_path_list:
+                return compiled_metadata_details
+        # Otherwise, extract & compile auth metadata details (for each path)
+        compiled_passport_auth_issuers = []
+        compiled_bearer_auth_issuers = []
+        compiled_supported_types = []
+        for authz in authz_path_list:
+            authz_metadata_details = authz_metadata[authz]
+            # Reduce issuer list(s) to a sorted set to address potential duplicates..
+            if "passport_auth_issuers" in authz_metadata_details:
```
k-burt-uch, line 316: See above comment about the default passport issuer

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+                compiled_passport_auth_issuers = sorted(
+                    set(authz_metadata_details["passport_auth_issuers"]).union(
+                        set(compiled_passport_auth_issuers)
+                    )
+                )
+            if "bearer_auth_issuers" in authz_metadata_details:
```
k-burt-uch, line 324: See the above comment about the default bearer issuer

indexd/drs/blueprint.py
```diff
@@ -129,9 +258,145 @@ def create_drs_uri(did):
...
+            missing_error_guids.append(i)
+    # Check the authz for each returned object:
+    resolved_count = 0
+    for doc in docs:
+        # Resolve individual
+        guid = doc["did"]
+        if guid in missing_error_guids:
+            continue
+        try:
+            authorization = resolve_single_object_auth(object_id=guid)
```
k-burt-uch: This method has 
```
    except IndexNoRecordFound as err:
        raise IndexNoRecordFound(err)
```

Which will be caught by the exception block below and added to unexpected_error_guids. I think you need another except block to catch the IndexNoRecordFound and add it to the missing guids list. 

Most likely won't happen since we're checking the existence above, but its worth handling the raised exception on its own to avoid confusion.

indexd/drs/blueprint.py
```diff
@@ -99,10 +153,85 @@ def list_drs_records():
...
+    "/ga4gh/drs/v1/objects", methods=["POST"], provide_automatic_options=False
+)
+def get_drs_objects():
+    """Returns DRS objects for each provided DRS object id.
+    Expects 'bulk_object_ids' in request body"""
+    data = flask.request.get_json(force=True)
+    # Exit with malformed error return if missing object id
+    if "bulk_object_ids" not in data:
+        return handle_user_error("Request is malformed. Missing bulk object ids.")
+    ret = resolve_bulk_object_auth(id_list=data["bulk_object_ids"])
```
SpencerAxelrod, line 191: POST /objects returns full DrsObjects not just authorizations. Use `get_bulk()`, convert each guid with `indexd_to_drs()`, and place full DrsObjects in `resolved_drs_objects` key in response. Missing IDs in `unresolved_drs_objects`.

indexd/drs/blueprint.py
```diff
@@ -210,29 +496,70 @@ def indexd_to_drs(record, expand=False):
                 bundle_object.pop("contents", None)
             drs_object["contents"].append(bundle_object)
 
-    # access_methods mapping
     if "urls" in record:
-        drs_object["access_methods"] = []
+        # Add access_method key-value pair (required to append url info)
+        if "access_methods" not in record:
```
SpencerAxelrod: this should be `if "access_methods" not in drs_object:`

indexd/drs/blueprint.py
```diff
@@ -210,29 +496,70 @@ def indexd_to_drs(record, expand=False):
...
                     "access_id": location_type,
-                    "region": "",
+                    "available": available.get(location, True),
+                    "region": region.get(location, ""),
                 }
             )
 
+    # Add authorization metadata to access_methods if record object id NOT a bundle
+    # Auth metadata is optional for bundles
+    # if 'bundle_data' not in record:
```
SpencerAxelrod: I know we went over this a lot, but I actually think we want:

```python
if form == "object":
    did = record["did"]
    authorizations = resolve_single_object_auth(object_id=did)
    if drs_object.get("access_methods"):
        # only put auth inside actual AccessMethod entries
        for entry in drs_object["access_methods"]:
            entry["authorizations"] = authorizations
    # If no URLs -> leave access_methods key out (or empty list)
```

tests/test_drs.py
```diff
@@ -408,6 +434,441 @@ def test_drs_service_info_no_information_configured(
...
+            "unresolved": n_404,
+        },
+        "unresolved_drs_objects": unresolved_drs_objects,
+        "resolved_drs_objects": resolved_drs_objects,
+    }
+    all_dids = dids_200 + dids_404
+    return all_dids, expected_json, expected_404_dict
+
+
+def test_bulk_post(client, user, combined_default_and_single_table_settings):
```
SpencerAxelrod, line 841: here too POST /objects needs  assert DrsObjects are returned not just authorizations

indexd/drs/blueprint.py
nrandorf: Sorting returns the set union as a list and makes testing consistent when comparing outputs.

indexd/drs/blueprint.py
nrandorf: updated

indexd/drs/blueprint.py
nrandorf: updated

tests/test_drs.py
nrandorf: updated

indexd/drs/blueprint.py
nrandorf: Public cases have been addressed in recent updated with default handling.

nrandorf: Multiple resource paths addressed.

indexd/drs/blueprint.py
nrandorf: Updated!

indexd/drs/blueprint.py
nrandorf: Updated, resolution is within indexd_to_drs.

indexd/drs/blueprint.py
nrandorf: My thought process was that if the guid was found earlier, but then not found later, it's an unexpected error (i.e. not a simple index not found error since we check for this before the try block).

k-burt-uch: That makes sense. Could you add a quick comment to the exception block explaining this is intentional?

indexd/drs/blueprint.py
nrandorf: Outdated, I think the refactored code handles the error cases more appropriately than this previous version.

indexd/drs/blueprint.py
nrandorf: Updated!

indexd/drs/blueprint.py
nrandorf: The passport info has been updated in the default/default test settings. The passport and bearer issuers are handled/appended the same way. I think this addresses the concern raised in this comment, but let me know if I misunderstood.

indexd/drs/blueprint.py
nrandorf: Are you saying that if a default bearer info is present, then we should force the passport info to be empty? If not, I think the comment above is addressed in the latest updates.

indexd/drs/blueprint.py
```diff
@@ -129,9 +272,151 @@ def create_drs_uri(did):
...
+            ]
+            compiled_metadata_details["passport_auth_issuers"] = [
+                blueprint.default_passport_issuer
+            ]
+            return compiled_metadata_details
+        # If auth path doesn't match config auth data, return empty authz data
+        if authz_path_list_length == 1:
+            path = authz_path_list[0]
+            config_path_list = list(authz_metadata.keys())
+            if path not in config_path_list:
```
k-burt-uch: Can you clarify the code flow here? If the authz is not in the authorization metadata, then we return the defaults. But if it is in the authorization metadata, then we rely on the loop on 317. Is that correct?

nrandorf: Yes, that is correct. Made an update to improve readability.

indexd/drs/blueprint.py
nrandorf: No problem, adding a comment for improved readability.

tests/test_drs.py
```diff
@@ -27,21 +29,27 @@ def generate_presigned_url_response(did, status=200, **query_params):
 
 def get_doc(
     has_version=True,
-    urls=list(),
+    urls=None,
```
jbarno, line 32: oooh, good catch.

tests/test_drs.py
```diff
@@ -355,51 +402,43 @@ def test_get_drs_with_encoded_slash(
...
-            "name": "CTDS",
-            "url": "https://fictitious-commons.io",
-        },
-    }
-
     res = client.get("/ga4gh/drs/v1/service-info")
 
     assert res.status_code == 200
-    assert res.json == expected_info
+    data = res.json
```
jbarno, line 412: I would have kept the dictionary equality assertion the same and just changed expected_info dictionary. Truly a nothing burger though. Same test excluding the maxBulkRequestLength being a  `<` instead of `==`

jbarno (dismissed):
lgtm

indexd/drs/blueprint.py
```diff
@@ -198,10 +487,36 @@ def indexd_to_drs(record, expand=False):
...
+    region = {}
+    urls_metadata = record.get("urls_metadata", {})
+    for url, meta in urls_metadata.items():
+        if isinstance(meta, dict) and meta.get("region"):
+            region[url] = meta["region"]
+
+    if "urls" in record and record["urls"]:
+        for url in record["urls"]:
+            if url.startswith("s3://") and url not in region:
+                bucket_name = url.split("/")[2]
```
MichaelLukowski, line 502: should we `urlparse` for this instead of split and index? Just in case a s3 url is malformed.

MichaelLukowski (dismissed):
Looks good!

One small nitpick but it should be fine

indexd/drs/blueprint.py
nrandorf: Good point about malformed handling. urlparse won't provide validation, but we can add in some conditional handling for malformed scenario (i.e. if the scheme is unexpected, we can skip the split and index entirely).

nrandorf: On further review, I think we can leave this as is because the existing code already avoids appending url data if the input is malformed. A match in the subsequent lookup_bucket_region() helper is required and if the match is not found, then region information for the malformed url defaults to an empty string.

jacob50231 (approved):

PR #454, opened 2026-08-10
fix(BDC-1250): Adds preferred_type for DRS 1.5

k-burt-uch:
Updates indexd DRS to allow for a configurable default preferred_type, allowing DRS Servers to set the precedence of credentials that should be tried by DRS Clients.

Updates indexd DRS to allow for a per-authz preferred_type, allowing DRS Servers to override the default preferred_type for specific authz resource paths.

### New Features
- Updates indexd DRS to allow for a configurable default preferred_type, allowing
    DRS Servers to set the precedence of credentials that should be tried by DRS
    Clients.
-  Updates indexd DRS to allow for a per-authz preferred_type, allowing DRS
    Servers to override the default preferred_type for specific authz resource
    paths.
### Deployment changes
- DEFAULT_PREFERRED_TYPE environment variable can be set to configure all authz resource paths to have a preferred supported type. 
- DRS_AUTHORIZATION_METADATA environment variable is updated to allow each resource path to have a preferred support type. See https://github.com/uc-cdis/gen3-helm/pull/671

### DIFF
--- a/indexd/drs/blueprint.py
+++ b/indexd/drs/blueprint.py
@@ -5,6 +5,7 @@ import flask
 import json
 from cdislogging import get_logger
 import copy
+
 from indexd.errors import AuthError, AuthzError
 from indexd.errors import UserError
 from indexd.index.errors import NoRecordFound as IndexNoRecordFound
@@ -288,7 +289,7 @@ def resolve_single_object_auth(object_id: str) -> dict:
         ret = blueprint.index_driver.get_with_nonstrict_prefix(object_id)
         authz_path_list = ret["authz"]
         authz_metadata = copy.deepcopy(blueprint.drs_authorization_metadata)
-
+        preferred_type = blueprint.default_preferred_type
         # Define default (empty) metadata details to return
         compiled_metadata_details = {
             "drs_object_id": object_id,
@@ -338,15 +339,23 @@ def resolve_single_object_auth(object_id: str) -> dict:
             compiled_metadata_details["bearer_auth_issuers"] = sorted(
                 compiled_bearer_auth_issuers
             )
+            if "preferred_type" in authz_metadata_details:
+                preferred_type = authz_metadata_details["preferred_type"]
 
         # Update supported_types
         compiled_supported_types = []
-        if compiled_passport_auth_issuers:
-            compiled_supported_types.append("PassportAuth")
-        if compiled_bearer_auth_issuers:
-            compiled_supported_types.append("BearerAuth")
+        if preferred_type == "PassportAuth":
+            if compiled_passport_auth_issuers:
+                compiled_supported_types.append("PassportAuth")
+            if compiled_bearer_auth_issuers:
+                compiled_supported_types.append("BearerAuth")
+        else:
+            if compiled_bearer_auth_issuers:
+                compiled_supported_types.append("BearerAuth")
+            if compiled_passport_auth_issuers:
+                compiled_supported_types.append("PassportAuth")
 
-        compiled_metadata_details["supported_types"] = sorted(compiled_supported_types)
+        compiled_metadata_details["supported_types"] = compiled_supported_types
         return compiled_metadata_details
     except IndexNoRecordFound as err:
         raise IndexNoRecordFound(err)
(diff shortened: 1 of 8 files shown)


tests/test_drs.py
```diff
@@ -74,6 +74,33 @@ def get_bundle(client, user, has_description=True):
...
+    assert authorizations["bearer_auth_issuers"] == [
+        "https://gen3.datacommons.io",
+        "sample_url",
+    ]
+    assert authorizations["drs_object_id"] == entry["id"]
+    assert authorizations["passport_auth_issuers"] == [
+        "https://ras/foo/bar",
+        "https://ras/foo/bar/bar",
+    ]
+    assert authorizations["supported_types"] == expected_supported_types or [
```
SpencerAxelrod: this will never fail without parens: `(expected_supported_types or ["PassportAuth", "BearerAuth"])`

SpencerAxelrod: also, I tried adding the parens and got `AssertionError: assert ['PassportAuth', 'BearerAuth'] == ['BearerAuth', 'PassportAuth']` so fix that however

k-burt-uch: Good catch. I've added parenthesis and updated the call to the assert helper for test_drs_list to use the correct expected supported types.

indexd/drs/blueprint.py
```diff
@@ -338,15 +339,23 @@ def resolve_single_object_auth(object_id: str) -> dict:
             compiled_metadata_details["bearer_auth_issuers"] = sorted(
                 compiled_bearer_auth_issuers
             )
+            if "preferred_type" in authz_metadata_details:
+                preferred_type = authz_metadata_details["preferred_type"]
```
SpencerAxelrod, line 343: `authz_path_list` can be randomly ordered I think, so if a record has multiple authz paths, this wont be deterministic. We could make sure Passport always wins ties over Bearer? Or just do longest/most specific path wins.

indexd/drs/blueprint.py
```diff
@@ -338,15 +339,23 @@ def resolve_single_object_auth(object_id: str) -> dict:
...
 
         # Update supported_types
         compiled_supported_types = []
-        if compiled_passport_auth_issuers:
-            compiled_supported_types.append("PassportAuth")
-        if compiled_bearer_auth_issuers:
-            compiled_supported_types.append("BearerAuth")
+        if preferred_type == "PassportAuth":
```
SpencerAxelrod, line 347: Do we have some kind of Validation to make to guard against Typos or lowercase like `preferred_type: "passport_awth"` lol

indexd/drs/blueprint.py
k-burt-uch: Yeah this is a good point. Multi-authz path support isn't a "real" use case so I don't think it'll be a problem if we leave it as is. It isn't like `acl` where we have things like `admin` alongside the other tags.

indexd/drs/blueprint.py
k-burt-uch: No, but the worst case is it'll default to preferring BearerAuth which will maintain continuity of service, even in passport enabled commons. 

Either way this should happen at the gen3-helm level to prevent deployments. I don't think we have jsonschema validation set up there though, so that's a bigger task.

tests/test_drs.py
```diff
@@ -902,3 +1021,63 @@ def test_bulk_auth_options_malformed_error(
     # Define expected results
     assert res_1.status_code == 400
     assert res_1.json["msg"] == "Request is malformed. Missing bulk object ids."
+
+
+def test_preferred_type(client, user, combined_default_and_single_table_settings):
+    """
+    Tests that DEFAULT_PREFERRED_TYPE is applied to projects that are not configured to have their own preferred_type
+      In default_test_settings.py:
+        - /gen3/programs/a/projects/d is configured to use the default as the preferred_type.
```
SpencerAxelrod: I think this is referring to either `/gen3/programs/a/projects/b` or `/gen3/programs/c/projects/d`

k-burt-uch: Whoops, that's what I get for copy and pasting. I"ve fixed this is both tests

tests/default_test_settings.py
```diff
@@ -27,6 +27,7 @@
 
 CONFIG["DRS_AUTHORIZATION_METADATA"] = {
     "/gen3/programs/a/projects/b": {
+        "preferred_type": "PassportAuth",
         "supported_types": ["BearerAuth", "PassportAuth"],
```
SpencerAxelrod: Since we are fully just deriving `supported_types` should remove from the default configs?

k-burt-uch: supported_types isn't even valid for DRS_AUTHORIZATION_METADATA so they cna get removed from here.

SpencerAxelrod (dismissed):

SpencerAxelrod (approved):

